PR #632, opened 2026-07-23
feat(eslint-markdown)!: add `no-trailing-heading-punctuation` rule

tooth-is-silver:
## Summary

This pull request implements the rule proposed in [eslint/markdown#682](https://github.com/eslint/markdown/issues/682), which was redirected to [`lumirlumir/npm-eslint-markdown`](https://github.com/lumirlumir/npm-eslint-markdown) because ESLint core does not accept new stylistic rules.

The rule reimplements markdownlint's MD026 behavior without depending on markdownlint.

## Details

- Check trailing punctuation in ATX, closed ATX, and Setext headings.
- Use `.,;:!。，；：！` as the default prohibited characters.
- Support custom characters through the `punctuation` option. An empty string disables checking.
- Ignore the trailing semicolon of HTML entity references such as `&copy;` and the trailing colon of GitHub emoji codes such as `:smile:`.
- Remove trailing punctuation and any immediately preceding whitespace with autofix.
- Register the rule in the `all` and `stylistic` configurations.
- Add rule tests, documentation, and migration mappings for markdownlint and remark-lint.

### References

The following similar pull requests were reviewed for repository conventions:

- [#587 - Add `no-consecutive-blank-line`](https://github.com/lumirlumir/npm-eslint-markdown/pull/587): stylistic rule, autofix, configuration, and migration conventions
- [#444 - Create `require-heading-id`](https://github.com/lumirlumir/npm-eslint-markdown/pull/444): heading rule, test, and documentation conventions
- [#547 - Add `no-double-punctuation`](https://github.com/lumirlumir/npm-eslint-markdown/pull/547): punctuation matching and option conventions

### Breaking change

The new rule is enabled in the `all` and `stylistic` configurations. Users of these configurations may receive new lint errors for existing headings.

- `npm test`
  - All 95 tests for `no-trailing-heading-punctuation` passed.
- `npm run lint`
- `npm run build`

All commands completed successfully.

## Resolved Issues

Closes eslint-ossca/eslint-ossca-2026#15

## Checklist


lumirlumir:
@tooth-is-silver Thanks for working on this new rule implementation! I’ll take a look at it over the weekend.

packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts
```diff
@@ -0,0 +1,134 @@
...
+ * @author Ga eun Lee(tooth-is-silver)
+ */
+
+// --------------------------------------------------------------------------------
+// Import
+// --------------------------------------------------------------------------------
+
+import { URL_RULE_DOCS } from '../core/constants.js';
+import type { RuleModule } from '../core/types.js';
+import escapeStringRegexp from '../core/utils/escape-string-regexp.js';
```
lumirlumir:
```suggestion
import { escapeStringRegexp } from '../core/utils/index.js';
import { URL_RULE_DOCS } from '../core/constants.js';
import type { RuleModule } from '../core/types.js';
```

I think we can use `index.js` to import the `escapeStringRegexp` helper here and move it to the top of the import list to comply with the existing convention.

packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts
```diff
@@ -0,0 +1,134 @@
...
+        },
+        additionalProperties: false,
+      },
+    ],
+
+    defaultOptions: [
+      {
+        punctuation: '.,;:!。，；：！',
+      },
+    ],
```
lumirlumir, line 81: I personally prefer not passing the options as a string, such as `".,;:"`. Instead, they could be passed as an array, such as `[".", ",", ";", ":"]`, for clarity and to follow the existing convention.

Also, the rule schema can be restricted by using `minLength: 1` and `maxLength: 1`, as shown below:

https://github.com/lumirlumir/npm-eslint-markdown/blob/432359e04bc81336a32bb148a866a89a09097f8d/packages/eslint-markdown/src/rules/no-double-punctuation.ts#L58-L59

packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts
```diff
@@ -0,0 +1,134 @@
...
+    dialects: ['commonmark', 'gfm'],
+  },
+
+  create(context) {
+    const { sourceCode } = context;
+    const [{ punctuation }] = context.options;
+
+    if (punctuation === '') {
+      return {};
+    }
```
lumirlumir: Instead of returning early when `punctuation` is passed as an empty string (`''`), I think we could restrict the rule schema so that it does not accept an empty array.

That way, the rule won’t silently run when an empty string (`''`) is passed. Instead, ESLint will throw an error when an empty array is provided.

lumirlumir (changes requested):
Sorry for the delay, and thanks for getting started on this!

I’ve left a few comments. I think this rule may need a few more iterations after this review, but it seems like a good start. 🙂

---

Just some references for me to look up in the future:

- https://github.com/DavidAnson/markdownlint/blob/main/doc/md026.md
- https://github.com/DavidAnson/markdownlint/blob/main/lib/md026.mjs
- https://github.com/DavidAnson/markdownlint/blob/main/helpers/helpers.cjs#L37

packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts
tooth-is-silver: Thanks for the detailed review!

Agreed. I'll switch to an array of single characters.

I'll apply `minLength: 1`, `maxLength: 1` and `uniqueItems` to match `no-double-punctuation`, and change the default to `['.', ',', ';', ':', '!', '。', '，', '；', '：', '！']`, which is the same set of characters as the current default.
I'll update the docs and tests accordingly.

packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts
tooth-is-silver: That makes sense. I used the early return to follow MD026, where an empty value disables the rule.

I'll add `minItems: 1` and remove the early return. I'll update the docs and tests accordingly.

lumirlumir (commented):
Thanks for the updates! Overall, it looks really good to me.

I’ll add some more test cases to prevent regressions and cover some tricky edge cases in this PR, and then I’ll merge it ASAP 🍻

tooth-is-silver:
Thanks for the fix! I checked the code and the new tests.

I found one more edge case.
When an entity is immediately followed by a punctuation character, both get removed.
`# Copyright &copy;.` becomes `# Copyright &copy`.

I think we could check if the `;` where the match starts closes a real entity, and start the match one character later if it does.

- It only needs to look at a few characters before the match.
- No new dependency, since `@eslint/markdown` already pulls in `decode-named-character-reference`.
- It also catches cases we miss now, like telling `&notreal;` apart from a real entity.

What do you think?

lumirlumir:
> `# Copyright &copy;.` becomes `# Copyright &copy.`

It’s kind of funny that `markdownlint` also doesn’t handle this case. 🙂 

image

However, it seems noteworthy, so I’ll find an easy way to resolve it.

lumirlumir:
[`decode-named-character-reference`](https://npmx.dev/package/decode-named-character-reference#user-content-when-should-i-use-this) mentions “Never use this,” and this package also couldn’t resolve the issue with gemoji patterns such as `# Heading :smile:.`, which still produce false positives. I’ve chosen another approach. (Done in https://github.com/lumirlumir/npm-eslint-markdown/pull/632/commits/20d07efcf83b2e92cf6c862eb3886439a79087b0)

image

lumirlumir (approved):
LGTM. Thanks for the contribution!

PR #640, opened 2026-08-02
feat(eslint-markdown): add no-heading-like-paragraph rule

Gaic4o:
## Summary

CommonMark ATX headings support at most six `#` characters. As a result, `####### Installation` is rendered as a paragraph rather than a heading, even though it can easily look like a heading in the source and be written by mistake.

This PR adds the `no-heading-like-paragraph` rule to detect such cases.

## Details

* The rule detects `paragraph` nodes whose raw source starts with seven or more `#` characters followed by a CommonMark ATX heading delimiter (` `, `\t`, a line ending, or EOF).
* It checks the raw source text instead of the AST `text` value. Otherwise, intentionally escaped or character-referenced inputs such as `\####### Foo` and `&#35;###### Foo` could be reported incorrectly.
* The rule provides two suggestions instead of an automatic fix:

  * Replace the leading hashes with `######` to make it a valid heading.
  * Escape the first `#` to keep it as a paragraph.
* An automatic fix is not provided because the intended outcome cannot be determined from the number of hashes alone.
* The rule is configured with `recommended: false` and `stylistic: false`, so it is included only in the `all` config.
* Continuation lines within a paragraph are not checked. This limitation is documented in the rule documentation.

This PR also adds the rule documentation and the corresponding `remark-lint` migration mapping.

`npm run test`, `npm run build`, and `npm run lint` all pass.

The rule is tested in both CommonMark and GFM modes, including valid and invalid cases as well as the outputs of both suggestions.

## Resolved Issues

Related to eslint/markdown#700

## Checklist


lumirlumir:
Thanks for sending a PR! I’ll review it in a few days.

lumirlumir:
@Gaic4o 

After giving it some more thought, I think it would be better to wait for consensus from the `@eslint/markdown` side rather than implement this rule directly in this repository.

Since the rule reports an anti-pattern rather than enforcing a stylistic preference, I’d prefer to reopen the issue in `@eslint/markdown`. If the team decide not to adopt it, we can then consider implementing it here.

Gaic4o:
> After giving it some more thought, I think it would be better to wait for consensus from the `@eslint/markdown` side rather than implement this rule directly in this repository.
> 
> Since the rule reports an anti-pattern rather than enforcing a stylistic preference, I’d prefer to reopen the issue in `@eslint/markdown`. If the team decide not to adopt it, we can then consider implementing it here.

Sorry for the late reply. Understood. I'll close this ticket for now and reopen the issue that was previously closed in `eslint/markdown` so we can continue the discussion there. Thank you!

PR #647, opened 2026-08-10
feat(eslint-markdown)!: add `consistent-heading-style` rule

### LINKED ISSUE #646: Suggestion: Implement `consistent-heading-style`
## Overview

This rule is based on [markdownlint `MD003`](https://github.com/DavidAnson/markdownlint/blob/main/doc/md003.md) and enforces a consistent heading style throughout a document.

```markdown
<!-- Incorrect -->
# ATX style H1

## Closed ATX style H2 ##

<!-- Correct -->
# ATX style H1

## ATX style H2
```

## Proposed Changes

Add a `consistent-heading-style` rule.

- Reimplement the same behavior as markdownlint's `MD003` without depending on markdownlint.
- Pass the same test cases as markdownlint.
- Support the `style` option with the following values:
  - `atx`
  - `atx-closed`
  - `consistent`
  - `setext`
  - `setext-with-atx`
  - `setext-with-atx-closed`
- Use `consistent` as the default style.
- Set `stylistic: true` so the rule is included in the stylistic configuration.
- Add autofix support to automatically convert headings to the configured style.
- Add the rule implementation and tests under `packages/eslint-markdown/src/rules/`.
- Export the new rule from `packages/eslint-markdown/src/rules/index.js`.
- Add documentation for the new rule under `website/docs/rules/`.
- Update the migration guide in `docs/get-started/migration.md`.
- Run `node scripts/build-config.mjs` to rebuild the configurations.

## Prior Art

- markdownlint: [`MD003` - Heading style](https://github.com/DavidAnson/markdownlint/blob/main/doc/md003.md)
- remark-lint: [`remark-lint-heading-style`](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-heading-style)

## Notes

- This is a breaking-change PR. Use the `feat(eslint-markdown)!` prefix in the PR title.
- Before writing the rule, tests, or documentation, review at least three existing rules with similar names or behavior and follow their conventions.
- Refer to [`CONTRIBUTING.md`](https://github.com/lumirlumir/npm-eslint-markdown/blob/main/CONTRIBUTING.md) for contribution guidelines.

tooth-is-silver:
## Summary

This pull request adds the `consistent-heading-style` rule proposed in #646.

The rule reimplements markdownlint's MD003 behavior without depending on markdownlint. It enforces a consistent heading style across ATX, closed ATX, and Setext headings.

## Details

- Support `atx`, `atx-closed`, `consistent`, `setext`, `setext-with-atx`, and `setext-with-atx-closed`.
- Use `consistent` as the default style.
- Use the first heading style when `consistent` is configured.
- Report level 3 through 6 headings with the strict `setext` style, following MD003.
- Autofix heading styles when the conversion preserves the Markdown structure.
- Preserve inline Markdown and existing line endings during fixes.
- Avoid unsafe fixes for empty headings, nested ATX-to-Setext conversions, and multiline Setext-to-ATX conversions.
- Register the rule in the `all` and `stylistic` configurations.
- Add tests, rule documentation, and migration mappings for markdownlint and remark-lint.

### References

The following rules were reviewed for repository conventions:

- [`consistent-unordered-list-style`](https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/rules/consistent-unordered-list-style.ts): consistent style and configuration conventions
- [`consistent-thematic-break-style`](https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/rules/consistent-thematic-break-style.ts): style option and autofix conventions
- [`no-trailing-heading-punctuation`](https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts): heading, test, and documentation conventions

The rule behavior is based on:

- [markdownlint MD003 documentation](https://github.com/DavidAnson/markdownlint/blob/main/doc/md003.md)
- [markdownlint MD003 implementation](https://github.com/DavidAnson/markdownlint/blob/main/lib/md003.mjs)
- [`remark-lint-heading-style`](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-heading-style#readme)

### Breaking change

The new rule is enabled in the `all` and `stylistic` configurations. Users of these configurations may receive new lint errors for existing headings.

- `npm run test`
  - All 2,243 tests passed, including 87 new tests for `consistent-heading-style`.
- `npm run lint`
- `npm run build`

All commands completed successfully.

## Checklist

Closes #646


tooth-is-silver:
I considered how to handle heading style mismatches that cannot be fixed safely.

I chose to report them without an autofix. For example:
```md
Title
=====

### Subheading
```
When `consistent` selects Setext from the first heading, the level 3 heading is reported but not fixed because Setext cannot represent level 3.
Following MD003, the mismatch is still reported. Autofix is skipped when the conversion could change the Markdown structure.

There is also a UI issue on the documentation site.
image

The rule reports the full Setext range, but the **Twoslash** renderer shows the red underline only on the first line.

Please let me know if you have any thoughts on these decisions or prefer a different approach.

lumirlumir:
Thanks for sending the PR! I’ll review it within a few days.

packages/eslint-markdown/src/rules/consistent-heading-style.ts
```diff
@@ -0,0 +1,231 @@
...
+            if (currentHeadingStyle === 'setext') {
+              const isMultilineHeading =
+                headingContent.includes('\n') || headingContent.includes('\r');
+
+              if (!isMultilineHeading) {
+                const headingMarker = '#'.repeat(node.depth);
+
+                replacementText =
+                  expectedHeadingStyle === 'atx'
+                    ? `${headingMarker} ${headingContent}`
```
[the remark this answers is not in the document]
tooth-is-silver: Confirmed. I found the case I missed and will add support for it.

packages/eslint-markdown/src/rules/consistent-heading-style.ts
```diff
@@ -0,0 +1,231 @@
...
+                isPreviousLineBlank;
+
+              if (canConvertToSetext) {
+                const lineEnding = sourceCode.text.match(/\r\n|\r|\n/)?.[0] ?? '\n';
+
+                const underlineMarker = node.depth === 1 ? '=' : '-';
+
+                replacementText = `${headingContent}${lineEnding}${underlineMarker.repeat(
+                  headingContent.length,
+                )}`;
```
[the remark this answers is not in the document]
tooth-is-silver: Confirmed. I found the case I missed and will add a safety check and regression test.

packages/eslint-markdown/src/rules/consistent-heading-style.ts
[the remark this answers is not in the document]
tooth-is-silver: Fixed in the latest commit bc38793.
I added tests for single and multiple trailing hashes. 👍

packages/eslint-markdown/src/rules/consistent-heading-style.ts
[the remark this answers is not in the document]
tooth-is-silver: Addressed in a89c321

Supporting every case would require partially reimplementing [CommonMark block parsing](https://spec.commonmark.org/0.31.2/#blocks-and-inlines)(ATX headings, blockquotes, lists, fenced code blocks, thematic breaks, link reference definitions, and HTML blocks).
Please let me know if you have any other suggestions.

packages/eslint-markdown/src/rules/consistent-heading-style.ts
```diff
@@ -0,0 +1,245 @@
...
+        type: 'object',
+        properties: {
+          style: {
+            enum: [
+              'consistent',
+              'atx',
+              'atx-closed',
+              'setext',
+              'setext-with-atx',
+              'setext-with-atx-closed',
```
lumirlumir:
```ts
const HEADING_STYLE = [
  'consistent',
  'atx',
  'atx-closed',
  'setext',
  'setext-with-atx',
  'setext-with-atx-closed',
] as const;
```

I think we can add the helper constant above and use it as a single source of truth for both `RuleOptions` and the rule schema.

For example, the current `RuleOptions` could be changed to the following:

```ts
type HeadingStyle = (typeof HEADING_STYLE)[number];
type RuleOptions = [{ style: HeadingStyle }];
```

I’d also prefer to remove `type HeadingStyle = 'atx' | 'atx-closed' | 'setext';` and inline it into the return types of the two helper functions, `getCurrentHeadingStyle` and `getExpectedHeadingStyle`.

For example:

```ts
function getExpectedHeadingStyle(
  style: Exclude<HeadingStyle, 'consistent'>,
  depth: number,
): 'atx' | 'atx-closed' | 'setext' {
  if (style === 'setext-with-atx') {
    return depth <= SETEXT_MAX_DEPTH ? 'setext' : 'atx';
  }

  if (style === 'setext-with-atx-closed') {
    return depth <= SETEXT_MAX_DEPTH ? 'setext' : 'atx-closed';
  }

  return style;
}
```

packages/eslint-markdown/src/rules/consistent-heading-style.ts
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Matches heading content that may start a block-level construct after conversion to Setext.
+ * @see https://spec.commonmark.org/0.31.2/#blocks-and-inlines
+ */
+const potentialBlockStartRegex = /^(?:>|(?:[-+*]|\d{1,9}[.)])(?:[ \t]|$))/u;
+
+function getCurrentHeadingStyle(text: string): HeadingStyle {
+  // ATX headings occupy one line, so a heading node containing a line ending is Setext.
+  if (text.includes('\n') || text.includes('\r')) {
```
lumirlumir: I think the heading type can be determined by checking whether the node spans multiple lines with `loc.start.line !== loc.end.line`, using the location information derived from `sourceCode#getLoc`, instead of checking for `\n` and `\r` directly.

Since the parser determines the line-ending format, checking `\n` and `\r` directly here seems hard-coded.

---

Also, since this function is only used in one place, removing the helper function and inlining its logic directly into `create(context)` would be clearer.

packages/eslint-markdown/src/rules/consistent-heading-style.ts
```diff
@@ -0,0 +1,245 @@
...
+            const [, contentEndOffset] = sourceCode.getRange(lastChildNode);
+
+            const headingContent = sourceCode.text.slice(
+              contentStartOffset,
+              contentEndOffset,
+            );
+
+            if (currentHeadingStyle === 'setext') {
+              const isMultilineHeading =
+                headingContent.includes('\n') || headingContent.includes('\r');
```
lumirlumir: Same here. I think we can use the node’s position information instead. 🙂

lumirlumir (changes requested):
Thanks for working on this! I think it may need a few iterations, like the ones we made in https://github.com/lumirlumir/npm-eslint-markdown/pull/632, but let’s take it one step at a time. :)

packages/eslint-markdown/src/rules/consistent-heading-style.test.ts
```diff
@@ -0,0 +1,525 @@
...
+      options: [{ style: 'atx-closed' }],
+    },
+    {
+      name: 'ATX heading containing an escaped trailing hash is not closed',
+      code: '## Heading \\#',
+      options: [{ style: 'atx' }],
+    },
+    {
+      // `markdownlint` MD003 only checks ATX and Setext tokens, so HTML heading elements are ignored.
+      // @see https://github.com/DavidAnson/markdownlint/blob/3f1f479322e863a53e56c94b01266b9785cd3bfd/lib/md003.mjs#L15
```
lumirlumir:
```suggestion
      // see: https://github.com/DavidAnson/markdownlint/blob/3f1f479322e863a53e56c94b01266b9785cd3bfd/lib/md003.mjs#L15
```

Small suggestion: The `@see` syntax only has meaning in JSDoc comments, not in `//` comments :)

packages/eslint-markdown/src/rules/consistent-heading-style.test.ts
```diff
@@ -0,0 +1,525 @@
...
+
+  invalid: [
+    {
+      name: '`consistent` style uses the first ATX heading',
+      code: '# Heading\n\n## Heading ##\n\nHeading\n-------',
+      output: '# Heading\n\n## Heading\n\n## Heading',
+      errors: [
+        {
+          messageId: 'style',
+          line: 3,
```
lumirlumir, line 181: Can we add the full `line`, `column`, `endLine`, and `endColumn` properties to all `invalid` test cases? This will help us detect future regressions.

lumirlumir (changes requested):

packages/eslint-markdown/src/rules/consistent-heading-style.test.ts
```diff
@@ -0,0 +1,525 @@
...
+      name: 'Document without headings',
+      code: 'Paragraph\n\n---',
+    },
+    {
+      name: '`consistent` style with one ATX heading',
+      code: '# Heading',
+    },
+    {
+      name: '`consistent` style with ATX headings at every level',
+      code: '# H1\n\n## H2\n\n### H3\n\n#### H4\n\n##### H5\n\n###### H6',
```
lumirlumir: Since we’re not testing the CR or CRLF line-ending detection logic, I’d prefer using multiline strings with backticks instead of repeatedly using `\n\n` for clarity.

For example:

```js
`
# H1

## H2

### H3

#### H4

##### H5

###### H6
`
```

---

FYI, this also applies to all other test cases that use the `\n` pattern more than twice.

packages/eslint-markdown/src/rules/consistent-heading-style.ts
tooth-is-silver: Updated in [d33b82b](https://github.com/lumirlumir/npm-eslint-markdown/pull/647/commits/d33b82b6d0d216fa3dfec726fac46303ed9ed057)

packages/eslint-markdown/src/rules/consistent-heading-style.ts
tooth-is-silver: 👍 Updated in [21e659d](https://github.com/lumirlumir/npm-eslint-markdown/pull/647/commits/21e659d1cd39456514f2b2e08370fa941e8c1414)

packages/eslint-markdown/src/rules/consistent-heading-style.ts
tooth-is-silver: Updated in [21e659d](https://github.com/lumirlumir/npm-eslint-markdown/pull/647/commits/21e659d1cd39456514f2b2e08370fa941e8c1414)

packages/eslint-markdown/src/rules/consistent-heading-style.test.ts
tooth-is-silver: Updated in 7545d6f

packages/eslint-markdown/src/rules/consistent-heading-style.test.ts
tooth-is-silver: I missed that. Thanks for catching it! Updated in a972b86.

packages/eslint-markdown/src/rules/consistent-heading-style.test.ts
tooth-is-silver: Updated in 33b3240

lumirlumir (commented):
I investigated the implementation, and the autofix logic under `heading:exit` is actually quite tricky.

- First, the two regex patterns in the helper don’t handle backslash escapes. For example, `## hi \##` isn’t a closed ATX heading, and `# \>` isn’t a blockquote when auto-fixed.
- Second, `lineEnding = sourceCode.text.match(/\r\n|\r|\n/)?.[0] ?? '\n';` relies on a heuristic that assumes the first line terminator is used throughout the document.
- There may be other issues as well.

I think we could consider autofixing later rather than implementing it all in a single PR, or limit the autofix to cases where the behavior is very straightforward.

What do you think, @tooth-is-silver?

tooth-is-silver:
I agree.
Handling every autofix edge case would require a lot of effort.
Should we remove autofix entirely, or limit it to specific cases?

lumirlumir:
@codex review

packages/eslint-markdown/src/rules/consistent-heading-style.ts
```diff
@@ -0,0 +1,377 @@
...
+                  const [, lastChildNodeEndOffset] = sourceCode.getRange(
+                    node.children[node.children.length - 1],
+                  );
+
+                  yield fixer.removeRange([nodeStartOffset, firstChildNodeStartOffset]);
+
+                  yield fixer.replaceTextRange(
+                    [lastChildNodeEndOffset, nodeEndOffset],
+                    `\n${getSetextMarker(node.depth).repeat(lastChildNodeEndOffset - firstChildNodeStartOffset)}`,
+                  );
```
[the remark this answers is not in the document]
lumirlumir: I think it's fine for the first version. We can revisit it later.

packages/eslint-markdown/src/rules/index.ts
```diff
@@ -38,6 +39,7 @@ export default {
   'consistent-code-style': consistentCodeStyle,
   'consistent-delete-style': consistentDeleteStyle,
   'consistent-emphasis-style': consistentEmphasisStyle,
+  'consistent-heading-style': consistentHeadingStyle,
```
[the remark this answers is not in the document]
lumirlumir: It's not something required. It's optional.

lumirlumir (approved):
Thanks for the contribution!

PR #663, opened 2026-08-21
docs(eslint-markdown): add JSDoc to rule option types

tooth-is-silver:
## Summary

Add JSDoc descriptions to `RuleOptions` and their properties.

This prepares the option types for a future change that will provide type information and editor hints in JavaScript and TypeScript configuration files.

## Details

- Added JSDoc descriptions to `RuleOptions` and their properties across all rule files.
- Kept the existing option types and runtime behavior unchanged.

- `npm run lint`
- `npm run test`

Both commands passed.

## Resolved Issues

N/A

## Checklist


tooth-is-silver:
@codex review

packages/eslint-markdown/src/rules/allow-heading.ts
```diff
@@ -17,9 +17,15 @@ import type { RuleModule } from '../core/types.js';
 // --------------------------------------------------------------------------------
 
 export interface HeadingOptions {
+  /** Markdown source patterns allowed at this heading level. */
   allow: (RegExp | string)[];
+  /** Markdown source patterns disallowed at this heading level. */
```
lumirlumir:
```suggestion
  /** 
   * Markdown source patterns allowed at this heading level. 
   * @default [/.*/u]
   */
allow: (RegExp | string)[];
  /** 
   * Markdown source patterns disallowed at this heading level. 
   * @default []
   */
```

It would be wonderful if we could add an `@default` section using JSDoc syntax. I hope this can be applied to every rule. 🙂

lumirlumir (changes requested):
Thanks for submitting a PR! I’ve left a single comment on it. :)

packages/eslint-markdown/src/rules/allow-heading.ts
tooth-is-silver: Added the `@default` tags for all rule options in [1db02f1](https://github.com/lumirlumir/npm-eslint-markdown/pull/663/commits/1db02f1039ddc604e428bd2be111df1a375f2c21). 👍

packages/eslint-markdown/src/rules/allow-heading.ts
```diff
@@ -17,9 +17,21 @@ import type { RuleModule } from '../core/types.js';
 // --------------------------------------------------------------------------------
 
 export interface HeadingOptions {
+  /**
+   * Markdown source patterns allowed at this heading level.
```
lumirlumir:
```suggestion
   * Allowed heading patterns. Only headings matching at least one pattern are allowed.
```

https://github.com/lumirlumir/npm-eslint-markdown/blob/bcab65fa33effdcd85b901e9b036793a947e73e6/website/docs/rules/allow-heading.md?plain=1#L132

I hope the JSDoc description can be the same as the one in the documentation, as it will serve as a type hint for end users in the future. I’d like to keep it consistent with the existing documentation rather than mix the two up.

Some rule options have multiline explanations, and it’s fine to use the full multiline description here.

The same applies to the other places.

lumirlumir (changes requested):

packages/eslint-markdown/src/rules/allow-heading.ts
tooth-is-silver: Updated in [29e2fc3](https://github.com/lumirlumir/npm-eslint-markdown/pull/663/commits/29e2fc318b7b5b2d4fa863df5eb99a4befe23b90)!

I included the essential option details and omitted the `For example` sections from the type hints.

packages/eslint-markdown/src/rules/allow-heading.ts
```diff
@@ -17,9 +17,21 @@ import type { RuleModule } from '../core/types.js';
 // --------------------------------------------------------------------------------
 
 export interface HeadingOptions {
+  /**
+   * Allowed heading patterns. Only headings matching at least one pattern are allowed.
+   * @default [new RegExp('.*', 'u')]
```
lumirlumir: It’s kind of funny that `*/` in a regex pattern is recognized as the end of the comment, so we can’t use the regex literal form here. 🤣

I’ll add a small comment next to it to clarify that this is actually a regex literal.

lumirlumir:
@codex review

packages/eslint-markdown/src/rules/allow-heading.ts
```diff
@@ -17,9 +17,21 @@ import type { RuleModule } from '../core/types.js';
 // --------------------------------------------------------------------------------
 
 export interface HeadingOptions {
+  /**
+   * Allowed heading patterns. Only headings matching at least one pattern are allowed.
+   * @default [new RegExp('.*', 'u')] // But as a regex literal.
```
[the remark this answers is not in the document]
lumirlumir: See https://github.com/lumirlumir/npm-eslint-markdown/pull/663#discussion_r3852907555

lumirlumir (commented):
Thanks for the PR!

PR #670, opened 2026-08-23
feat(eslint-markdown): add auto-fix and `override` option to `no-irregular-dash`

### LINKED ISSUE #666: Add auto-fix and `override` option to `no-irregular-dash`
## Description

The `md/no-irregular-dash` rule reports irregular Unicode dash characters, but it does not currently provide an auto-fix. Add an auto-fix that replaces each reported character with a configured replacement.

The replacements should come from a frozen default map, following the pattern used by `code-lang-shorthand`, instead of being hard-coded in the fixer. This keeps the default behavior explicit and lets users customize individual replacements.

## Motivation

The rule documentation recommends the standard ASCII hyphen-minus (`-`, `\u002D`) in place of Unicode lookalikes. Users currently have to replace every reported character manually, even when the intended replacement is unambiguous.

Auto-fixing these reports would make the rule easier to adopt across an existing documentation set. A configurable map is also needed because some projects may prefer multi-character ASCII replacements, such as `--` for an en dash or `---` for an em dash.

## Proposed Solution

Define a read-only replacement map in `no-irregular-dash.ts`, analogous to `langShorthandMap` in `code-lang-shorthand.ts`. The map should contain every character that should be treated as an irregular dash. To preserve the rule's current documented policy and keep the automatic behavior predictable, each default replacement should be the ASCII hyphen-minus.

```ts
const irregularDashMap: Readonly<Record<string, string>> = Object.freeze({
  '\u2010': '-',
  '\u2011': '-',
  '\u2012': '-',
  '\u2013': '-',
  '\u2014': '-',
  '\u2015': '-',
  '\u2043': '-',
  '\u2212': '-',
  '\u23af': '-',
  '\u2e3a': '-',
  '\u2e3b': '-',
  '\u30fc': '-',
  '\ufe58': '-',
  '\ufe63': '-',
  '\uff0d': '-',
});
```

Add an `override` rule option with the following shape and default value:

```ts
type RuleOptions = [
  {
    allow: string[];
    override: Record<string, string>;
    skipCode: boolean | string[];
    skipInlineCode: boolean;
  },
];

// In meta.defaultOptions
override: {},
```

Merge `override` after the default map so a user-supplied value takes precedence, matching the approach in `code-lang-shorthand`:

```ts
const mergedIrregularDashMap = {
  ...irregularDashMap,
  ...override,
};
```

For example, users should be able to preserve a project's preferred ASCII style:

```js
'md/no-irregular-dash': ['error', {
  override: {
    '\u2013': '--',
    '\u2014': '---',
  },
}]
```

Mark the rule as `fixable: 'code'`. For each report, use `fixer.replaceTextRange([startOffset, endOffset], replacement)` with the value from the merged map.

Existing behavior must remain intact:

- A character included in `allow` is neither reported nor fixed.
- Characters covered by `skipCode` or `skipInlineCode` are neither reported nor fixed.
- Report locations and the `noIrregularDash` message remain unchanged.
- Without `override`, every reported irregular dash is fixed to `-`.

### Keep the default map conservative

Some dash characters have reasonable multi-character ASCII approximations. For example, the Unicode Standard notes that an em dash (`U+2014`) is commonly represented as `--` on a typewriter, while `U+2E3A` and `U+2E3B` represent dashes wider than an em dash. However, there is no single ASCII replacement convention that applies to every project or language. See [Unicode Standard, Section 6.2](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-6/).

For that reason, `--` and `---` should be explicit, opt-in replacements supplied through `override`, rather than default values. Keeping every default replacement as `-` also remains consistent with the rule's documented goal of normalizing characters to the ASCII hyphen-minus.

### Warn when an override can change Markdown structure

Multi-character replacements can change the parsed Markdown structure. In CommonMark, a line containing three or more hyphens can become a thematic break, and `---` below a paragraph can become a Setext heading underline. See the CommonMark sections on [thematic breaks](https://spec.commonmark.org/0.31.2/#thematic-breaks) and [Setext headings](https://spec.commonmark.org/0.31.2/#setext-headings).

The documentation for `override` should warn about this risk, especially for `---` and longer replacements. These values remain useful, but users should opt into them with knowledge of their Markdown semantics.

## Acceptance Criteria

- `no-irregular-dash` declares `meta.fixable` as `code`.
- A frozen `Readonly<Record<string, string>>` contains a default replacement for every character matched by the updated `irregularDashRegex`.
- Every default replacement is `-`; alternative `--` or `---` replacements require an explicit `override`.
- The `override` option is included in `RuleOptions`, `meta.schema`, and `meta.defaultOptions`.
- User-provided replacements take precedence over default replacements.
- Tests cover default fixes, overridden fixes, multiple occurrences, and the interaction with `allow`, `skipCode`, and `skipInlineCode`.
- Rule documentation describes the new `override` option, auto-fix behavior, and the Markdown syntax risk of multi-hyphen replacements.

Grit03:
## Summary

Add an auto-fix to `no-irregular-dash`, with an `override` option to customize each replacement.

## Details

- Replaced `irregularDashRegex` with a frozen `irregularDashMap`, following `code-lang-shorthand`, and derived the regex from its keys.
- Added `override`, merged after the default map so user values take precedence. Defaults stay `-`, so `--` and `---` are opt-in.
- Marked the rule `fixable: 'code'` and fixed each report with `fixer.replaceTextRange`.
- `allow`, `skipCode`, `skipInlineCode`, report locations, and the message are unchanged.
- Documented `override` with an example and a warning that multi-hyphen replacements can turn into a thematic break or a Setext heading, and added a `Fix` section describing the auto-fix.

Added test cases for:

- The default fix for every character in the map.
- `override` replacing a mapped character.
- A partially overridden map, where the remaining characters still fix to `-`.
- Multiple occurrences of an overridden character.
- `override` combined with `allow`, `skipCode`, and `skipInlineCode`.

`npm run test` and `npm run lint` both passed.

## Resolved Issues

closes: #666 

## Checklist


website/docs/rules/no-irregular-dash.md
```diff
@@ -195,6 +205,31 @@ Examples of **correct** code for this rule:
 
 When specified, specific irregular dash characters are allowed if they match one of the characters in this array. This is useful for ignoring certain irregular dashes that are intentionally used in the document.
 
+### `override`
+
+> Type: `Record<string, string>` / Default: `{}`
+
+An object where the **key** is an irregular dash character and the **value** is the string that replaces it. Only the characters listed in [Rule Details](#rule-details) can be used as keys.
```
Grit03, line 212: I wrote this line as the intended usage, but the schema does not actually enforce it. `override` uses `additionalProperties: { type: 'string' }`, so a key outside the map still passes validation, and since `irregularDashRegex` is built from the `irregularDashMap` keys, such a key never matches and is silently ignored at runtime.

Would you rather restrict the keys in the schema so this line holds true?

```js
override: {
  type: 'object',
  properties: Object.fromEntries(
    Object.keys(irregularDashMap).map(key => [key, { type: 'string' }]),
  ),
  additionalProperties: false,
},

Comment (tooth-is-silver) [reply website/docs/rules/no-irregular-dash.md:212]: All supported irregular dash characters currently use the same default replacement, `-`.
Because of this, I think an object map adds unnecessary structure, and it would be better to represent the supported characters as an array.

For example:
```ts
const irregularDashes = [
  '\u2010',
  '\u2011',
  '\u2012',
  '\u2013',
  '\u2014',
  '\u2015',
  '\u2043',
  '\u2212',
  '\u23af',
  '\u2e3a',
  '\u2e3b',
  '\u30fc',
  '\ufe58',
  '\ufe63',
  '\uff0d',
] as const;

const irregularDashPattern = `[${irregularDashes.join('')}]`;

const irregularDashRegex = new RegExp(irregularDashPattern, 'gu');
``` 

 The override schema could then use `patternProperties`, following the same approach as ESLint's [no-multi-spaces rule](https://github.com/eslint/eslint/blob/c27bc926e496985eb7911c09eb60914b2e4b5d0f/lib/rules/no-multi-spaces.js#L53-L61):
 
 ```ts
override: {
  type: 'object',
  patternProperties: {
    [`^[${irregularDashes.join('')}]$`]: {
      type: 'string',
    },
  },
  additionalProperties: false,
},

...
fix(fixer) {
  return fixer.replaceTextRange(
     [startOffset, endOffset],
     override[irregularDash] ?? '-',
  );
},
```
This approach would allow us to:
- Avoid generating schema properties from object keys.
- Avoid using an object map when every default replacement is `-`.
- Derive the supported key type directly from the array.
- Follow the patternProperties approach used in ESLint core.

What do you think about this approach?

Comment (lumirlumir) [reply website/docs/rules/no-irregular-dash.md:212]: > Would you rather restrict the keys in the schema so this line holds true?

Actually, `code-lang-shorthand` was originally designed to accept additional property keys, but in this case, I think we can restrict the object’s property keys to only the one already defined.

This would also resolve the comment mentioned in https://github.com/lumirlumir/npm-eslint-markdown/pull/670#discussion_r3838589687.

https://github.com/lumirlumir/npm-eslint-markdown/blob/a492589d7b89449c53ebcfe378e7bb2ab1b999aa/packages/eslint-markdown/src/rules/code-lang-shorthand.ts#L46

---

I think what @tooth-is-silver suggested is a good idea, but I’d prefer an `irregularDashMap` variable in the helper, built from the `irregularDashes` array, to follow the existing convention of `code-lang-shorthand`’s `langShorthandMap`.

**Review by lumirlumir — CHANGES_REQUESTED** [review]
Thanks for submitting a PR! Could you please take a look at the merge conflict?

File: packages/eslint-markdown/src/rules/no-irregular-dash.ts [hunk 2]
```diff
@@ -134,6 +172,13 @@ export default {
             },
 
             messageId: 'noIrregularDash',
+
+            fix(fixer) {
+              return fixer.replaceTextRange(
+                [startOffset, endOffset],
+                mergedIrregularDashMap[irregularDash],
+              );
```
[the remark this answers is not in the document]
lumirlumir: I think we can add a caution callout to the `override` section of the documentation mentioning that the case described above can cause an infinite loop.

Or, if there’s a clear way to resolve this problem, it would be helpful to use that approach instead.

packages/eslint-markdown/src/rules/no-irregular-dash.ts
```diff
@@ -53,6 +79,12 @@ export default {
             },
             uniqueItems: true,
           },
+          override: {
+            type: 'object',
+            additionalProperties: {
+              type: 'string',
+            },
```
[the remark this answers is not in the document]
lumirlumir: See https://github.com/lumirlumir/npm-eslint-markdown/pull/670#discussion_r3853791151

website/docs/rules/no-irregular-dash.md
Grit03: Thanks for the reviews!

The dashes are now defined `irregularDashes` array and derived `irregularDashMap` and `irregularDashRegex` from it, keeping the list as the single source of truth. (954b3a4)

And I restricted the `override` keys to those same dashes with `additionalProperties: false`, so an unsupported key is now a config error instead of being silently ignored. (976dce1)

packages/eslint-markdown/src/rules/no-irregular-dash.ts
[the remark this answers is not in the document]
Grit03: > Or, if there’s a clear way to resolve this problem, it would be helpful to use that approach instead.

If the configurable map isn't meant to support replacing an irregular dash with another irregular dash, what if we just disallow `irregularDashPattern` in `override` values?

```ts
override: {
  type: 'object',
  properties: Object.fromEntries(
    irregularDashes.map(irregularDash => [
      irregularDash,
      {
        type: 'string',
        pattern: `^[^${irregularDashPattern}]*$`,
      },
    ]),
  ),
  additionalProperties: false,
},
```

That would take care of the circular case, and configs like `{ '—': '–' }` where the override ends up not applying anyway. What do you think, is that too strict?

lumirlumir (approved):
Thanks for the contribution! It looks so good to me. 🙂

PR #674, opened 2026-08-30
feat(eslint-markdown): add `no-shell-dollar` rule

### LINKED ISSUE #669: Suggestion: implement markdownlint's MD014 as `commands-show-output`
## Description

Add a new rule equivalent to markdownlint's [MD014](https://github.com/DavidAnson/markdownlint/blob/main/doc/md014.md), **Dollar signs used before commands without showing output**.

The rule reports a code block when every non-blank line is a shell command prefixed with a dollar sign (`$`). With `--fix`, it removes the prompt prefix.

### Incorrect

```sh
$ npm install
$ npm run build
```

### Correct

```sh
npm install
npm run build
```

A dollar-sign prompt is allowed when the block also contains command output:

```sh
$ npm --version
10.9.2
```

**`@eslint/markdown` does not currently have an equivalent rule, so this would not overlap with an existing rule.**

## Motivation

A dollar-sign prompt only adds useful information when a code block shows both input and output, where it distinguishes the command being typed from the output being printed.

When a block contains commands only, the prefix adds noise and makes the example less convenient to use:

- Readers need to strip each `$` before running the commands.
- Copy buttons attached to code blocks copy the `$` verbatim.

markdownlint enables MD014 by default, so this behavior is also relevant for projects migrating from markdownlint. MD014 is currently missing from the migration guide table.

## Proposed Solution

Add a rule named `commands-show-output`, with the markdownlint mapping documented as MD014 so the relationship remains clear in the migration guide.

### Detection

Follow the upstream MD014 behavior:

- A non-blank line counts as a command when it begins with optional indentation followed by a `$` prompt and whitespace.
- A bare `$`, and a `$` with no following whitespace such as `$npm install`, are treated as regular content.
- Blank lines are ignored.
- A code block is reported only when every remaining line is a prefixed command.
- Each prefixed command line is reported individually.
- Both fenced and indented code blocks are checked, regardless of language.

### Fix

For each reported command, remove the `$` and the whitespace immediately following it while preserving the indentation before the prompt.

For example:

```diff
-  $ npm install
+  npm install
```

### Rule metadata

Suggested metadata:

- `type: "problem"`
- `fixable: "code"`
- no options
- `dialects: ["commonmark", "gfm"]`

## Design Questions

There are two points where I would like feedback before opening the PR.

### 1. Recommended configuration

markdownlint enables MD014 by default, but this repository takes the opposite position for its own documentation: `.markdownlint.json` sets `"commands-show-output": false`, and the installation guide relies on the prompt prefix in every code group.

```sh [npm]
$ npm install -D eslint-markdown@latest eslint@latest
```

That style is a deliberate one, and the rule should not push projects away from it. My assumption is therefore that the rule ships in `all` only, with `recommended: false` and `stylistic: false`, and that this repository keeps the check disabled for its own documentation.

If you would rather the plugin not carry a rule that its own documentation opts out of, that is a reasonable place to close this proposal, and I would rather hear it before the PR than after.

### 2. Prompt characters

MD014 recognizes `$` only.

I considered making the prompt configurable, but that would make this rule a superset of the upstream behavior and make the migration mapping less exact. My current implementation therefore supports `$` only.

## Implementation Note

One implementation detail is that blockquote markers and list indentation affect source offsets.

The current implementation scans the code block's source lines in order and uses those offsets when reporting and fixing each command. This also avoids accidentally treating text on the opening fence line, such as:

````markdown
```sh $ ls
````

as part of the code block contents.

### PR DESCRIPTION
**Pull request by uncoolclub**
## Summary

Adds `md/no-shell-dollar`, based on markdownlint's [MD014](https://github.com/DavidAnson/markdownlint/blob/main/doc/md014.md). The design follows the decisions in #669.

The rule removes `$` prompts from code blocks that contain commands only. It leaves blocks that also show output unchanged.

The rule is not part of the `recommended` or `stylistic` config. Use `skipCode` to ignore selected code block languages.

### Detection

- A prompt matches `/^[ \t]*\$[ \t]+/`. Therefore, `$npm install` and a bare `$` stay unchanged.
- Blank lines are ignored and end a continued command.
- An odd number of backslashes at the end of a line continues the command. The next line has no prompt to report.
- Each prompt is reported only if every other non-blank line is a command or a continued line.
- The rule checks fenced and indented code blocks. It skips languages listed in `skipCode`.

### Differences from MD014

MD014 treats a continued line as output, so it passes this block:

````md
```sh
$ npm install \
    --save-dev eslint
```
````

This rule treats the second line as part of the command. It reports the first line and leaves the second line unchanged. A command followed by output still passes.

This rule also supports `skipCode`. Shell prompts such as `>` and `#` stay unchanged because they can mark a transcript or a root prompt.

## Scope

- `packages/eslint-markdown/src/rules/no-shell-dollar.ts`, `packages/eslint-markdown/src/rules/no-shell-dollar.test.ts`
- `packages/eslint-markdown/src/rules/index.ts`, `packages/eslint-markdown/src/configs/all.ts` (regenerated by the build)
- `website/docs/rules/no-shell-dollar.md`, `website/docs/get-started/migration.md`

## Validation

- `npm run test`: 2,252 tests pass, including 85 for this rule.
- `npm run build`: passes and regenerates `src/configs/all.ts`.
- `npm run lint`: passes.
- Manual ESLint checks cover blank lines, trailing spaces, escaped backslashes, and repeated text.

Tests also cover indented code blocks, blockquotes, list items, tab-separated prompts, and fence metadata that looks like a command.

## Notes

- Other prompt characters such as `>`, `PS>`, and `#` are out of scope for this version, as agreed in the issue.
- `skipCode` matches language identifiers as written. It cannot skip a code block without a language identifier.

Closes #669

Disclosure: I'm a participant of open source contribution program OSSCA

### CODE REVIEW

File: packages/eslint-markdown/src/rules/no-shell-dollar.ts [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+        const commandRanges: [number, number][] = [];
+        let previousLineContinues = false;
+
+        for (const codeLine of node.value.split('\n')) {
+          if (codeLine.trim() === '') {
+            previousLineContinues = false;
+            continue;
+          }
+
+          const codeLineOffset = nodeText.indexOf(codeLine, searchOffset);
```
uncoolclub, line 129: **Review focus: source offset mapping**

I chose to search forward through `nodeText`. Unlike `node.value`, it retains the fence and surrounding Markdown markers. Moving `searchOffset` after each match also maps repeated lines to the correct source range. The same approach works for fenced and indented blocks inside blockquotes and lists.

packages/eslint-markdown/src/rules/no-shell-dollar.ts
```diff
@@ -0,0 +1,150 @@
...
+
+          searchOffset = codeLineOffset + codeLine.length;
+
+          if (!previousLineContinues) {
+            const match = dollarCommandRegex.exec(codeLine);
+
+            if (!match?.groups) {
```
uncoolclub, line 136: **Review focus: all-or-nothing reporting**

I chose to treat the whole block as a transcript when a non-blank line is not a `$ ` command or continuation. This discards prompts collected earlier. It avoids a partial fix when the example includes command output.

uncoolclub (commented):
I left a few inline comments on the parts I am still thinking through. I would appreciate your feedback on them during the review. Thank you!

packages/eslint-markdown/src/rules/no-shell-dollar.ts
```diff
@@ -0,0 +1,150 @@
...
+              return;
+            }
+
+            const { indentation, prompt } = match.groups;
+            const startOffset = nodeStartOffset + codeLineOffset + indentation.length;
+
+            commandRanges.push([startOffset, startOffset + prompt.length]);
+          }
+
+          previousLineContinues =
```
uncoolclub, line 146: **Review focus: continuation semantics**

I chose an odd number of backslashes at the end of a line as the continuation boundary. This preserves escaped backslashes. A backslash followed by spaces or a blank line ends the continuation. This differs from MD014 so the rule can fix multiline commands without a shell parser.

website/docs/get-started/migration.md
```diff
@@ -44,6 +44,7 @@ This section is currently under construction and will be updated soon.
 | [`MD004` - Unordered list style](https://github.com/DavidAnson/markdownlint/blob/main/doc/md004.md#md004---unordered-list-style) :white_check_mark: | [`md/consistent-unordered-list-style`](../rules/consistent-unordered-list-style.md) |
 | [`MD010` - Hard tabs](https://github.com/DavidAnson/markdownlint/blob/main/doc/md010.md#md010---hard-tabs) :white_check_mark: | [`md/no-tab`](../rules/no-tab.md) |
 | [`MD012` - Multiple consecutive blank lines](https://github.com/DavidAnson/markdownlint/blob/main/doc/md012.md#md012---multiple-consecutive-blank-lines) :white_check_mark: | [`md/no-consecutive-blank-line`](../rules/no-consecutive-blank-line.md) |
+| [`MD014` - Dollar signs used before commands without showing output](https://github.com/DavidAnson/markdownlint/blob/main/doc/md014.md#md014---dollar-signs-used-before-commands-without-showing-output) :white_check_mark: | [`md/no-shell-dollar`](../rules/no-shell-dollar.md) |
```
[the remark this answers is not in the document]
uncoolclub: ☑️ 25bb76b

website/docs/get-started/migration.md
lumirlumir, line 1: Could you also add a migration guide for `remark-lint` as well?

lumirlumir (commented):
Thanks for working on this! I’ll take a look in a few days. 🙂

lumirlumir:
> ```sh
> $ npm install \
>     --save-dev eslint
> ```

Praise: `markdownlint` and `remark-lint` don’t report this case, but I think reporting it reflects the rule’s use case even better. Nice find!

website/docs/get-started/migration.md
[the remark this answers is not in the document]
lumirlumir: I think marking it as `:white_check_mark:` is fine here, since checking `\` is more accurate and this should be treated as a false positive or negative bug in both `markdownlint` and `remark-lint`. 🙂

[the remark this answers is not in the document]
lumirlumir: Updated it in https://github.com/lumirlumir/npm-eslint-markdown/pull/674/changes/12e9fbc0edc818636f7af0aa726a6d09873c6d5b

website/docs/get-started/migration.md
lumirlumir: Updated it in https://github.com/lumirlumir/npm-eslint-markdown/pull/674/changes/12e9fbc0edc818636f7af0aa726a6d09873c6d5b

packages/eslint-markdown/src/rules/no-shell-dollar.ts
```diff
@@ -0,0 +1,150 @@
...
+    return {
+      code(node) {
+        if (node.lang && skipCode.includes(node.lang)) {
+          return;
+        }
+
+        const [nodeStartOffset] = sourceCode.getRange(node);
+        const nodeText = sourceCode.getText(node);
+
+        let searchOffset = /^[`~]/.test(nodeText) ? nodeText.indexOf('\n') + 1 : 0;
```
lumirlumir: Instead of using a `` /^[`~]/ `` regex pattern here, I think we could promote the following utility to `core/utils` and reuse that logic.

Also, adding a brief unit test for the following utility would be helpful.

https://github.com/lumirlumir/npm-eslint-markdown/blob/0417a586cbe581c908d7af0b8e1a285600f5c9c2/packages/eslint-markdown/src/rules/consistent-code-style.ts#L60-L73

lumirlumir: Markdown’s valid line endings are CR, LF, and CRLF. So, directly using `\n` can’t correctly recognize line endings that use only CR.

A fenced code block always starts from the second line of the code block, so I think we can use that already calculated information rather than finding the offset again.

Also, adding test cases for the CR line ending pattern would be helpful to avoid regression.

packages/eslint-markdown/src/rules/no-shell-dollar.ts
```diff
@@ -0,0 +1,150 @@
...
+
+        const commandRanges: [number, number][] = [];
+        let previousLineContinues = false;
+
+        for (const codeLine of node.value.split('\n')) {
```
lumirlumir: Same here: using `\n` alone can’t recognize CR line endings. A regression test here would also be helpful.

packages/eslint-markdown/src/rules/no-shell-dollar.ts
```diff
@@ -0,0 +1,150 @@
...
+          if (codeLine.trim() === '') {
```
lumirlumir: `String#trim` also treats VF, FF, NBSP, and other non-space/tab characters as valid whitespace. Since CommonMark only treats spaces and tabs as valid whitespace, this is where the spec deviation occurs.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#white_space

I think we could use the following blank-line detection utility instead for a more accurate, spec-conformant approach:

https://github.com/lumirlumir/npm-eslint-markdown/blob/0417a586cbe581c908d7af0b8e1a285600f5c9c2/packages/eslint-markdown/src/core/utils/is-blank-line.ts#L24

A regression test for VF, FF, and NBSP here would be helpful.

lumirlumir (changes requested):

uncoolclub:
@codex review

Focus on the two commits addressing the review feedback:

- `b419f12` promotes the code style detection from `consistent-code-style` to `core/utils/get-code-style.ts`
- `da35bfb` fixes CR line endings and uses the CommonMark blank line definition in `no-shell-dollar`

packages/eslint-markdown/src/rules/no-shell-dollar.ts
uncoolclub: ☑️ b419f12

Promoted it as `getCodeStyle` in `core/utils/get-code-style.ts` together with `CODE_STYLE` and the `CodeStyle` type, and added a unit test. `consistent-code-style` now imports it from there.

uncoolclub: ☑️ da35bfb, a8355de

Lines are now split with `/\r\n|[\r\n]/u`, and the CR / CRLF cases are covered by tests, including a backslash continuation across CRLF.

While adding the tab regression below I also replaced the offset search entirely. Each value line now maps to its source line (`nodeStartLine + 1 + index` for fenced blocks), so the rule reads the prompt from `sourceCode.lines` and no longer looks up the second-line offset at all.

packages/eslint-markdown/src/rules/no-shell-dollar.ts
uncoolclub: ☑️ da35bfb, a8355de

Same fix as above. The `indexOf` search this comment pointed at is gone in a8355de.

packages/eslint-markdown/src/rules/no-shell-dollar.ts
uncoolclub: ☑️ da35bfb

Switched to `isBlankLine` and added valid cases for a line holding only NBSP, VT, or FF so each stays content rather than a blank line.

uncoolclub:
Addressed the review in three commits.

- b419f12 promotes the code style detection to `core/utils/get-code-style.ts` with a unit test.
- da35bfb handles CR line endings and uses `isBlankLine` for the CommonMark blank line definition, with regression tests for CR, CRLF, NBSP, VT, and FF.
- a8355de fixes a bug I found while adding those tests. When list item indentation splits a tab, `Code#value` holds spaces the source line does not, so the `indexOf` search returned -1 and the fix removed part of the opening fence:

  ````md
  - ```sh
  	$ ls
    ```
  ````

  The rule now maps each value line to its source line and reads the prompt from `sourceCode.lines`. That also drops the offset bookkeeping and the second-line lookup you asked about.

Test count is 2,276 for the repo, and `tsc -b` and `eslint` pass. Could you take another look when you have time?

lumirlumir:
Could you revert all the comment line deletion in https://github.com/lumirlumir/npm-eslint-markdown/pull/674/changes/a8355deb77ccb7b27cc2ae0c56577e87b1c7cccd and resolve the merge conflict when you have a moment? I’d like to retain the comment line for future reference. 🙂

image

uncoolclub:
Done in d9ca675 and aab7058!

- d9ca675 restores the comment lines removed in a8355de: the two `NOTE` blocks in the test file, and the two comments in the rule file (the fence one now reads "so its opening fence is skipped", since the offset search it described is gone).
- aab7058 merges `main` and resolves the conflict in `consistent-code-style.ts`. It keeps the `currentCodeFenceChar` variable and the narrowed error range from #675, and only swaps `getCurrentCodeStyle` for `getCodeStyle`.

Repo tests, `tsc -b`, and `eslint` pass after the merge.

packages/eslint-markdown/src/rules/no-shell-dollar.test.ts
```diff
@@ -0,0 +1,475 @@
...
+
+import ruleTester from '../tests/rule-tester.js';
+import rule from './no-shell-dollar.js';
+
+// --------------------------------------------------------------------------------
+// Test
+// --------------------------------------------------------------------------------
+
+ruleTester('no-shell-dollar', rule, {
+  valid: [
```
tooth-is-silver, line 18: The tests pass both with and without the `^` anchor in `dollarCommandRegex`.
Add a valid case where the `$` sits in the middle of a line. That makes the purpose of the anchor clear.

```ts
 {
   name: 'Dollar sign followed by a space in the middle of a command',
   code: '```sh\necho a $ b\n```',
 },
```

packages/eslint-markdown/src/rules/no-shell-dollar.ts
```diff
@@ -0,0 +1,167 @@
...
+    return {
+      code(node) {
+        if (node.lang && skipCode.includes(node.lang)) {
+          return;
+        }
+
+        // An empty code block has no command to report, and its opening fence may be the only line.
+        if (node.value === '') {
+          return;
+        }
```
tooth-is-silver: nit: an empty code block is already skipped by the `isBlankLine` check in the loop below, so this early return does not change the result. Removing it does not fail any test. If it covers a case I am missing, add that to the comment. Otherwise remove it.

tooth-is-silver (changes requested):

packages/eslint-markdown/src/rules/no-shell-dollar.test.ts
uncoolclub: Good catch, the anchor was not pinned by any test. Added the valid case you suggested in fffbe53.

Without the `^` the fix rewrites `echo a $ b` to `echo a b`, so the anchor is part of the rule's contract rather than an implementation detail. MD014 anchors the same way (`/^(\s*)(\$\s+)/`).

packages/eslint-markdown/src/rules/no-shell-dollar.ts
uncoolclub: You are right, removed it in 3810651.

The guard predated the line based lookup. The old code asked `getIndexFromLoc` for the second line, which throws when an empty opening fence is the last line of the file, so the early return kept that input away from it. Line numbers are arithmetic now, so `isBlankLine` in the loop already covers it.

I kept that input as a valid test case so the guard is not needed again.

lumirbot (dismissed):
(approving test)

packages/eslint-markdown/src/core/utils/index.ts
lumirlumir, line 1: Could you take a look at the merge conflict? 

I've changed the exports to use named export instead of default export in https://github.com/lumirlumir/npm-eslint-markdown/pull/707, so I think this logic can be simplified.

packages/eslint-markdown/src/core/utils/get-code-style.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const CODE_STYLE = ['indent', 'fence-backtick', 'fence-tilde'] as const;
+
+/**
+ * Get the style of a code block.
+ * - NOTE: Only the first character is inspected, so a single character read from the start of a `Code` node is enough.
+ * @param str Text starting at the beginning of a `Code` node.
+ * @returns The code style.
+ */
+export default function getCodeStyle(str: string): CodeStyle {
```
lumirlumir:
```suggestion
export function getCodeStyle(str: string): CodeStyle {
```

It'd be nice if we can use named export here either 🙂

lumirlumir (changes requested):

packages/eslint-markdown/src/core/utils/index.ts
uncoolclub: Resolved in 1d295c5. The barrel now follows #707, so the explicit import and re-export list is gone:

```ts
export * from './escape-string-regexp.js';
export * from './get-code-style.js';
...
```

That also drops the separate `export type { CodeStyle }` line, since `export *` carries the type along.

While I was there I added the two missing export checks to `core/utils/index.test.ts` (`CODE_STYLE` and `getCodeStyle`), which the file keeps one per exported utility.

packages/eslint-markdown/src/core/utils/get-code-style.ts
uncoolclub: Applied in 1d295c5. `getCodeStyle` is a named export now, and its test imports it that way. `CODE_STYLE` and `CodeStyle` were already named, so the whole module is consistent with the rest of `core/utils`.

lumirlumir (approved):
Thanks for the contribution! 🎉

tooth-is-silver (approved):

PR #693, opened 2026-09-07
feat(eslint-markdown): add `skipMath` and `skipInlineMath` options to `no-irregular-whitespace`

### LINKED ISSUE #689: Suggestion: add `skipMath` and `skipInlineMath` options to `no-irregular-whitespace`
## Overview

Add `skipMath` and `skipInlineMath` options to the [no-irregular-whitespace](https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/rules/no-irregular-whitespace.ts) rule.

The rule scans the entire source and reports characters such as thin space (U+2009) and narrow no-break space (U+202F), including inside parsed math. Users who intentionally preserve these characters in math currently have to allow them throughout the document with `allow`. Math-specific exclusions would keep the checks active in prose.

Both options should be booleans defaulting to `true`, consistent with `skipCode` and `skipInlineCode`. They apply to `math` and `inlineMath` nodes respectively and require `languageOptions: { math: true }`.

### 1. Update the Rule

In `packages/eslint-markdown/src/rules/no-irregular-whitespace.ts`, add both options to `RuleOptions`, `meta.schema`, and `meta.defaultOptions`. Read them in `create()` and add:

```ts
math(node) {
  if (skipMath) skipRanges.push(sourceCode.getRange(node));
},

inlineMath(node) {
  if (skipInlineMath) skipRanges.push(sourceCode.getRange(node));
},
```

Reuse `SkipRanges` and the existing source scan. Keep `allow` and the checked character set unchanged. Use AST ranges rather than matching math delimiters manually. This rule currently reports diagnostics only; do not add autofixes.

### 2. Add Tests

In `packages/eslint-markdown/src/rules/no-irregular-whitespace.test.ts`, cover:

- With math parsing enabled, irregular whitespace inside both math forms is skipped by default and with explicit `true` options.
- Setting either option to `false` enables reports in the corresponding math form.
- The two options work independently, including both mixed combinations.
- The same characters in surrounding prose are still reported, unless permitted by `allow`.
- Existing `allow`, `skipCode`, and `skipInlineCode` behavior remains intact.
- With math parsing disabled, math delimiters alone do not exclude characters.

Use explicit Unicode escapes such as `\u2009` in fixtures, individually named tests, and the existing helper for both CommonMark and GFM.

### 3. Update the Documentation

Update `website/docs/rules/no-irregular-whitespace.md` with both options' types, defaults, enabled/disabled examples, and the math-parsing prerequisite. Ensure automatically checked math examples enable math parsing.

Document that parsed math will be excluded by default; setting both options to `false` preserves the previous behavior. These exclusions do not guarantee that a particular math renderer supports every excluded whitespace character.

Related: #685, #687.

## Validation

For a fresh clone, run `npm install` from the repository root before implementation or validation. After making the changes, run:

```sh
npm run test
npm run build
npm run lint
```

vhmns14:
## Overview

Closes #689.

Adds `skipMath` and `skipInlineMath` options to the `no-irregular-whitespace` rule in `eslint-markdown`, allowing users to exclude irregular whitespace (such as thin space `\u2009` and narrow no-break space `\u202F`) inside math blocks and inline math when math parsing is enabled (`languageOptions: { math: true }`).

## Proposed Changes

1. **Rule update** (`packages/eslint-markdown/src/rules/no-irregular-whitespace.ts`):
   - Added `skipMath: boolean` and `skipInlineMath: boolean` to `RuleOptions`.
   - Registered both as boolean properties in `meta.schema`.
   - Set default values to `true` in `meta.defaultOptions`.
   - Added `math(node)` and `inlineMath(node)` listeners in `create()` to record ranges into `skipRanges` when the corresponding option is enabled.
2. **Tests** (`packages/eslint-markdown/src/rules/no-irregular-whitespace.test.ts`):
   - Verified that with math parsing enabled, irregular whitespace inside both math forms is skipped by default and with explicit `true` options.
   - Verified that `skipMath: false` causes irregular whitespace inside math blocks to be reported.
   - Verified that `skipInlineMath: false` causes irregular whitespace inside inline math to be reported.
   - Verified that the two options work independently, including both mixed combinations.
   - Verified that whitespace in surrounding prose is still reported unless permitted by `allow`.
   - Verified that with math parsing disabled, math delimiters alone do not exclude characters.
3. **Documentation** (`website/docs/rules/no-irregular-whitespace.md`):
   - Documented each option's type (`boolean`), default (`true`), and behavior.
   - Added examples for options enabled and disabled.
   - Documented the requirement to enable math parsing via `languageOptions: { math: true }`.

## Validation

- `npm run test` (type check + 2,215 unit tests passed).
- `npm run build` (packages build + website VitePress build passed).
- `npm run lint` (eslint, prettier, editorconfig, markdownlint passed).

### DIFF
--- a/packages/eslint-markdown/src/rules/no-irregular-whitespace.test.ts
+++ b/packages/eslint-markdown/src/rules/no-irregular-whitespace.test.ts
@@ -78,6 +78,67 @@ console.log(\u200b'Hello World');
         },
       ],
     },
+    {
+      name: '`skipMath: true` default: math block should be skipped',
+      code: `$$
+\\int f(x)\\,dx = \\sum a_i\\u2009b_i
+$$`,
+      languageOptions: {
+        math: true,
+      },
+    },
+    {
+      name: '`skipInlineMath: true` default: inline math should be skipped',
+      code: '$f(x)\\u2009dx$',
+      languageOptions: {
+        math: true,
+      },
+    },
+    {
+      name: '`skipMath: true, skipInlineMath: true` explicit options: math regions should be skipped',
+      code: `$$
+\\int f(x)\\u202Fdx
+$$
+
+$a\\u2009b$`,
+      options: [
+        {
+          skipMath: true,
+          skipInlineMath: true,
+        },
+      ],
+      languageOptions: {
+        math: true,
+      },
+    },
+    {
+      name: '`skipMath: true, skipInlineMath: false` mixed options: math block is skipped',
+      code: `$$
+\\int f(x)\\u2009dx
+$$`,
+      options: [
+        {
+          skipMath: true,
+          skipInlineMath: false,
+        },
+      ],
+      languageOptions: {
+        math: true,
+      },
+    },
+    {
+      name: '`skipMath: false, skipInlineMath: true` mixed options: inline math is skipped',
+      code: '$f(x)\\u2009dx$',
+      options: [
+        {
+          skipMath: false,
+          skipInlineMath: true,
+        },
+      ],
+      languageOptions: {
+        math: true,
+      },
+    },
   ],
 
   invalid: [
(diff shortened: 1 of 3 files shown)


lumirlumir:
@codex review

lumirlumir (approved):
Thanks for the contribution!

PR #709, opened 2026-09-13
feat(eslint-markdown): add math skip options to `no-irregular-dash`

### LINKED ISSUE #687: Suggestion: add `skipMath` and `skipInlineMath` options to `no-irregular-dash`
## Overview

Add `skipMath` and `skipInlineMath` options to the [no-irregular-dash](https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/rules/no-irregular-dash.ts) rule in [eslint-markdown](https://github.com/lumirlumir/npm-eslint-markdown).

The rule currently supports excluding code blocks and inline code, but has no equivalent options for math. Characters such as the Unicode minus sign (`−`, U+2212) are checked by this rule and may be intentionally used in math expressions.

Both options should default to `true`, consistent with `skipCode` and `skipInlineCode`. They should apply to `math` and `inlineMath` nodes respectively, which require `languageOptions: { math: true }`.

### 1. Update the Rule

In `packages/eslint-markdown/src/rules/no-irregular-dash.ts`:

- Add `skipMath: boolean` and `skipInlineMath: boolean` to `RuleOptions`.
- Register both as boolean properties in `meta.schema`.
- Set both default values to `true` in `meta.defaultOptions`.
- Read the options in `create()` and add the following listeners:

```ts
math(node) {
  if (skipMath) skipRanges.push(sourceCode.getRange(node));
},

inlineMath(node) {
  if (skipInlineMath) skipRanges.push(sourceCode.getRange(node));
},
```

Reuse the existing `SkipRanges` logic. Identify math regions using AST nodes rather than matching delimiters with a regular expression. Keep the existing `allow` behavior unchanged.

### 2. Add Tests

In `packages/eslint-markdown/src/rules/no-irregular-dash.test.ts`, verify that:

- With math parsing enabled, irregular dashes inside math blocks and inline math are skipped by default and with explicit `true` options.
- `skipMath: false` causes irregular dashes inside math blocks to be reported.
- `skipInlineMath: false` causes irregular dashes inside inline math to be reported.
- The two options work independently.
- Irregular dashes outside excluded math regions are still reported unless permitted by `allow`.
- With math parsing disabled, the options do not exclude text merely enclosed in math delimiters.

### 3. Update the Documentation

Update `website/docs/rules/no-irregular-dash.md` with:

- Each option’s type (`boolean`), default (`true`), and behavior.
- Examples with the options enabled and disabled.
- The requirement to enable math parsing through `languageOptions: { math: true }`.

Note that existing users with math parsing enabled will stop receiving reports inside math regions by default. Setting both options to `false` preserves the previous behavior.

## Validation

For a fresh clone, run `npm install` from the repository root before implementation or validation.

After making the changes, run:

```sh
npm run test
npm run build
npm run lint
```

electrohyun:
## Summary

Adds `skipMath` and `skipInlineMath` options to avoid reporting intentional irregular dashes in math regions.

## Details

Added `skipMath` and `skipInlineMath` options with default values of `true`.

Used `math` and `inlineMath` AST node ranges to exclude math regions while preserving the existing `allow` and autofix behavior.

Added tests for default and explicit options, independent option behavior, surrounding text, `allow`, and disabled math parsing.

Updated the rule documentation with examples, configuration requirements, and previous-behavior guidance.

I implemented the rule based on the issue requirements and updated the documentation by referring to similar prior PRs. 

For the tests below, I tried to keep each case focused on a single behavior wherever possible. I would appreciate any feedback.

```
Valid:
- Irregular dash in a math block should be skipped by default
- Irregular dash in inline math should be skipped by default
- `skipMath: true` should skip an irregular dash in a math block
- `skipInlineMath: true` should skip an irregular dash in inline math
- `allow` should permit an irregular dash outside math regions when math parsing is enabled

Invalid:
- `skipMath: false` should report an irregular dash in a math block
- `skipInlineMath: false` should report an irregular dash in inline math
- `skipMath: true, skipInlineMath: false` should skip only the math block
- `skipMath: false, skipInlineMath: true` should skip only inline math
- Irregular dash outside skipped math regions should be reported
- Math delimiters should not skip an irregular dash when math parsing is disabled
- Inline math delimiters should not skip an irregular dash when math parsing is disabled
```

## Resolved Issues

Closes #687

## Checklist

### DIFF
--- a/packages/eslint-markdown/src/rules/no-irregular-dash.test.ts
+++ b/packages/eslint-markdown/src/rules/no-irregular-dash.test.ts
@@ -45,6 +45,22 @@ console.log(\u2014'Hello World');
 
 \`console.log(\u2014'Hello World')\``,
     },
+    {
+      name: 'Irregular dash in a math block should be skipped by default',
+      code: `$$
+x\u2212y
+$$`,
+      languageOptions: {
+        math: true,
+      },
+    },
+    {
+      name: 'Irregular dash in inline math should be skipped by default',
+      code: '$x\u2212y$',
+      languageOptions: {
+        math: true,
+      },
+    },
 
     // Options
     {
@@ -108,6 +124,44 @@ console.log(\u2014'Hello World');
           override: { '\u2014': '---' },
         },
       ],
+    },
+    {
+      name: '`skipMath: true` should skip an irregular dash in a math block',
+      code: `$$
+x\u2212y
+$$`,
+      options: [
+        {
+          skipMath: true,
+        },
+      ],
+      languageOptions: {
+        math: true,
+      },
+    },
+    {
+      name: '`skipInlineMath: true` should skip an irregular dash in inline math',
+      code: '$x\u2212y$',
+      options: [
+        {
+          skipInlineMath: true,
+        },
+      ],
+      languageOptions: {
+        math: true,
+      },
+    },
+    {
+      name: '`allow` should permit an irregular dash outside math regions when math parsing is enabled',
+      code: 'Prose\u2013text',
+      options: [
+        {
+          allow: ['\u2013'],
+        },
+      ],
+      languageOptions: {
+        math: true,
+      },
     },
   ],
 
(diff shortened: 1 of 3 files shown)


electrohyun:
This is my first contribution to this repository, so I would appreciate feedback on any part of the changes, not just the tests. Thank you.

lumirlumir (approved):
Thanks for the contribution!

PR #711, opened 2026-09-13
docs(website): fix CommonJS configuration examples

electrohyun:
image

## Summary

Since this is a small documentation fix, I'm opening a PR directly without a separate issue.

The CommonJS configuration examples fail because `require()` returns a module namespace object instead of the plugin. This PR adds `.default` to access the plugin.

[Reproduce](https://stackblitz.com/edit/stackblitz-starters-fa3hqpft?file=eslint.config.cjs)

## Details

image

I think `require('eslint-markdown')` looks cleaner but the current ESM-only export requires `.default`.

Adding `export { plugin as 'module.exports' };` to `index.ts` seems to avoid this, as shown above, though it doesn't look particularly clean either.

Is the current behavior intentional, or are there other considerations?

Confirmed that adding `.default` fixes the error in the reproduction above.

## Resolved Issues

No issues

## Checklist

### DIFF
--- a/website/docs/get-started/configurations.md
+++ b/website/docs/get-started/configurations.md
@@ -63,8 +63,8 @@ export default defineConfig([
 
 ```js [eslint.config.cjs]
 const { defineConfig } = require('eslint/config');
-const markdown = require('@eslint/markdown');
-const md = require('eslint-markdown'); // [!code ++]
+const markdown = require('@eslint/markdown').default;
+const md = require('eslint-markdown').default; // [!code ++]
 
 module.exports = defineConfig([
   {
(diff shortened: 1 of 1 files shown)


electrohyun:
This PR is a draft only because I can currently have only one non-draft pull request open in this repository, not because the work is still in progress. If everything looks good, please feel free to mark it as ready and proceed. Thank you!

lumirlumir (approved):
Thanks for catching this!

I searched for `export { plugin as 'module.exports' };`, and it seems that adding this feature could break existing users.

https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require

Image

After adding:

```js
export { plugin as 'module.exports' };
```

CommonJS consumers on Node.js 22.12+ would receive the plugin directly:

```js
const md = require('eslint-markdown');
```

As a result, existing usage such as the following would no longer work because `.default` would be `undefined`:

```js
const md = require('eslint-markdown').default;
```

There is also a compatibility concern: this package supports Node.js `^20.19.0`, but the `'module.exports'` ESM interop export is only supported starting with Node.js 22.12.0. This would cause the `require()` API shape to differ depending on the Node.js version.

Therefore, I’d like to keep the `.default` shape for now, until we drop support for Node.js `^20.19.0`. 🙂

PR #718, opened 2026-09-14
feat(eslint-markdown): add `no-multiple-link-space` rule

### LINKED ISSUE #708: New Rule: implement `markdownlint`'s `MD039`
## Description

Add a rule equivalent to markdownlint's [MD039](https://github.com/DavidAnson/markdownlint/blob/main/doc/md039.md), **Spaces inside link text**. It reports leading and trailing whitespace inside link text and removes it with `--fix`.

```md
<!-- Incorrect -->

[ ESLint ](https://eslint.org)

<!-- Correct -->

[ESLint](https://eslint.org)
```

MD039 is listed under `Rule` in the [`1.0.0` roadmap](https://github.com/lumirlumir/npm-eslint-markdown/issues/437), and `@eslint/markdown` has no equivalent: `no-space-in-emphasis` does not look at links.

## Proposed Rule

- `type: "layout"`, `fixable: "whitespace"`, no options, `dialects: ["commonmark", "gfm"]`
- `recommended: false`, `stylistic: false`, following what you decided for `no-shell-dollar` in #669
- Covers `link` and `linkReference`. Images stay out of scope, matching MD039
- Removes spaces and tabs only, leaving line breaks in place, also matching MD039

That last point differs from `consistent-inline-code-style`, which lists `\r` and `\n` in its patterns, so it is worth saying why. Removing a line break from link text can change the output. Text ending in two spaces before the break renders a `<br>`, which disappears once those spaces go. Text ending in a backslash before the break escapes the closing bracket, so removing only the newline destroys the link. Leaving breaks alone keeps the fix from changing how a document renders, and we can revisit it later with a policy of its own.

## Question

**Naming.** Both `no-space-in-link-text` and `consistent-link-text-style` fit the singular form you described in #669. The first matches `no-space-in-emphasis` in `@eslint/markdown`. The second matches `consistent-inline-code-style`, which is how MD038, the code span counterpart of this rule, was named here. I lean toward the first, since the rule rejects one construct rather than picking among consistent styles, but I'm happy to use whichever fits the conventions better.

I can open the PR once the name is settled.

uncoolclub:
## Prerequisites checklist

## AI acknowledgment

Claude Code (Claude Fable 5.1) wrote the rule, its tests and its docs from my design decisions. I reviewed every line.

I also put the branch through several rounds of review with Codex. Most of the fixes below came from it.

I checked each finding by running the fix and rendering both sides. One report was wrong. I note that at the end.

Adds a rule equivalent to markdownlint's [MD039](https://github.com/DavidAnson/markdownlint/blob/main/doc/md039.md), **Spaces inside link text**, which the [`1.0.0` roadmap](https://github.com/lumirlumir/npm-eslint-markdown/issues/437) lists and `@eslint/markdown` has no equivalent for.

The rule is named `no-multiple-link-space`, following your call in this thread.

The rule reports spaces and tabs at the start and end of link text and removes them with `--fix`. It covers `link` and `linkReference`; autolinks and images have no link text between brackets and are out of scope.

**The label range comes from the source text, not from child positions.** The parser leaves padding next to a line break outside the children, so `[a\n   ](u)` has its trailing spaces beyond the last child's range. The rule instead takes the opening bracket as the first character of the node and the closing bracket as the first one after the last child, which holds for escaped brackets, nested brackets and code spans.

**Four kinds of padding are left alone, because removing them would change more than spacing.**

| Left alone | Why |
| :-- | :-- |
| Line breaks | Two spaces before one form a hard break; a backslash before one escapes the closing bracket |
| Link text spanning lines | It can hold either of the above, or a blockquote marker opening each line |
| Trailing padding after an odd number of backslashes | The backslash would escape the closing bracket |
| A shortcut reference opening a list item, reading `x` or `X`, followed by whitespace | GFM would read `[x]` as a checked task list item and the link would disappear |

The last one is checked against the syntax tree rather than the text before the link, because a list marker can sit on an earlier line or behind another marker. A `definition` is skipped when looking for the paragraph a list item opens, since it renders nothing and the parser looks past it.

Padding is counted in a loop rather than matched with an end-anchored pattern, which backtracked quadratically: 80,000 spaces took 3,549ms before and 37ms after.

```console
$ npm run test
Test Files  40 passed (40)
     Tests  3042 passed (3042)

$ npm run build
$ npm run lint
```

Coverage stays at 100 percent for statements, branches, functions and lines.

The rule has 123 test cases. Beyond those I checked the fix by rendering both sides through micromark with GFM and comparing, over roughly sixty inputs: nested and lazy lists, every list marker, blockquotes inside list items and lists inside blockquotes, escaped and nested brackets, images and autolinks, reference links in all three forms, CR, LF and CRLF, and non-ASCII link text. The four exceptions above each came out of that sweep or out of review, and each has a regression test.

I also ran the documentation's `eslint-check` blocks against the built rule to confirm every example matches its label.

## Related Issues

Closes #708

**Whether the reporting scope should match MD039.** MD039 reports padding made of line breaks alone and then declines to fix it. This rule stays silent there instead, on the grounds that it should not report what it will not fix. That is the one behavioural difference from MD039, and it is why the migration row is marked partially compatible rather than fully. The other difference worth noting is that only spaces and tabs count as padding here, so a no-break space is left in place where MD039 would remove it.

**The task list exception applies in both dialects.** A rule cannot tell which dialect it runs under, and CommonMark has no task list items, so that one position keeps its padding there too. The alternative was declaring the rule GFM-only, which would take it away from CommonMark users for a problem that exists in both.

**One known limitation, which I believe is an upstream parser bug rather than something this rule creates.** micromark renders this as a checkbox inside a paragraph that sits outside the list:

```md
-
[x] d

[x]: /url
```

Adding a blank line after the marker, or using any label other than `x`, renders a link as expected. The rule produces valid Markdown there and the parser misreads it, so I did not add a fifth exception for it. Happy to if you would rather be defensive.


uncoolclub:
I've opened this as a draft since you mentioned in #708 that you'd like some time to settle on the rule name. I'm not requesting a review yet, but thought seeing the implementation, tests, and docs together might help with the naming discussion.

I've used `no-space-in-link-text` as a placeholder for now. The name only appears in six files and doesn't affect the logic, so it'll be straightforward to update once a name is settled on.

I'll mark this ready for review then. Any thoughts are welcome in the meantime, but there's no rush to take a look.

lumirlumir (commented):
Thanks for sending a PR!

Now, I think that `no-multiple-link-space` rule name would be better fit here, as we are now implementing `no-multiple-atx-heading-space` rule in another PR.

Also, if you could take a look at the merge conflict, it would be great!

packages/eslint-markdown/src/rules/no-multiple-link-space.ts
```diff
@@ -0,0 +1,260 @@
...
+
+export default {
+  meta: {
+    type: 'layout',
+
+    docs: {
+      description: 'Disallow spaces at the start and end of link text',
+      url: URL_RULE_DOCS('no-multiple-link-space'),
+      recommended: false,
+      stylistic: false,
```
lumirlumir, line 91:
```suggestion
      stylistic: true,
```

I think it's fine to mark this PR as breaking change, and set the `stylistic: true` 🙂

website/docs/rules/no-multiple-link-space.md
```diff
@@ -0,0 +1,117 @@
...
+
+No options are available for this rule.
+
+## Fix
+
+This rule fixes the link text by removing the spaces and tabs next to the brackets.
+
+## Prior Art
+
+- [`MD039` - Spaces inside link text](https://github.com/DavidAnson/markdownlint/blob/main/doc/md039.md#md039---spaces-inside-link-text)
```
lumirlumir, line 117: If there's a similar `remark-lint` rule about this rule implementation, could you please add one it in here and in the `migration.md`?

PR #720, opened 2026-09-15
fix(eslint-markdown): ignore exempt `a` elements in `require-link-title`

### LINKED ISSUE #719: Bug: ignore `<a>` elements without `href` in `require-link-title`
### What rule do you want to change?

`require-link-title`

### What change do you want to make?

Generate fewer warnings

### How do you think the change should be implemented?

A new default behavior

### Example code

```markdown
<a id="section"></a>
```

### What does the rule currently do for this code?

The rule reports that the `<a>` element is missing a title attribute.

[reproduce](https://stackblitz.com/edit/stackblitz-starters-6yjgiguu?file=repro.md)

(run `npx eslint repro.md`)

### What will the rule do after it's changed?

The rule will ignore `<a>` elements without an `href` attribute.

### Participation

- [x] I am willing to submit a pull request to implement this change.

### AI acknowledgment

- [x] No - I did not use AI to generate this issue report.
- [ ] Yes - substantive assistance (AI-generated non-trivial portions of code, tests, or documentation).

### Additional comments

I think `<a>` elements without an `href` may be outside the scope of this rule.

The [HTML Standard’s definition of the `<a>` element](https://html.spec.whatwg.org/dev/text-level-semantics.html#the-a-element) explains that an `<a>` element without an `href` represents a placeholder. 

> If the a element has no href attribute, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element's contents.

The section on [links created by `<a>` elements](https://html.spec.whatwg.org/dev/links.html#links-created-by-a-and-area-elements) states that `href` is not required and that elements without it do not create hyperlinks. In addition, all HTML examples in the rule documentation include an `href`.

> The href attribute on a and area elements is not required; when those elements do not have href attributes they do not create hyperlinks.

For these reasons, I think the rule should ignore `<a>` elements without an `href`. Disabling the rule would also stop checking actual hyperlinks. I would appreciate your thoughts on whether this behavior is intended.

electrohyun:
## Prerequisites checklist

## AI acknowledgment

- [x] No - I did not use AI for this contribution.
- [ ] Yes - substantive assistance (AI-generated non-trivial portions of code, tests, or documentation).

Updates `require-link-title` to ignore `<a>` elements without `href` or with `aria-hidden='true'`.

Updated `require-link-title` rule to ignore `<a>` elements without `href` or with `aria-hidden="true"`.

Added tests and documentation for the new behaviors.

I added the following test cases to cover the new behavior and its edge cases.

```
Valid:
- An `<a>` element without an `href` attribute should be ignored
- An `<a>` element with `aria-hidden="true"` should be ignored

Invalid:
- An `<a>` element with an empty `href` and no title should be reported
- An `<a>` element with `aria-hidden="false"` and no title should be reported
- A link without a title following ignored links should still be reported
```

## Related Issues

Fixes #719

Here is the previous implementation code.

image

Initially, I considered using a single `for` loop with separate `if` statements for `href`, `title`, and `aria-hidden`. However, because the attribute order is not guaranteed, the loop often has to check every attribute. 

For example, if `title` appears before `href`, it cannot break immediately.

Therefore, I chose the current approach.

| Current approach | Initial approach |
| --- | --- |
| Current implementation | Initial implementation |

The current approach first checks whether the element is in scope using `href`, then checks for conditions that prevent a report. 

It also allows the attribute scan to stop once it finds a valid `title` or `aria-hidden='true'`.

Please let me know if any changes are needed to the documentation or tests.


electrohyun:
Since this changes the behavior of an existing rule, I’ve updated the PR title to feat. Sorry for the confusion!

website/docs/rules/require-link-title.md
```diff
@@ -11,6 +11,8 @@ The rule examines all link elements in a Markdown document and reports any links
 - Link reference definitions: `[ref]: url "title"`
 - HTML link tags: `<a href="url" title="title">text</a>`
 
+HTML `<a>` elements without an `href` attribute and elements with `aria-hidden="true"` are ignored.
```
tooth-is-silver: Could you update this sentence to make it clear that both conditions apply to HTML `<a>` elements?

> HTML `<a>` elements without an `href` attribute or with `aria-hidden="true"` are ignored.

electrohyun: @tooth-is-silver That makes sense to me, so I’ve applied the change. Thanks for the review!

tooth-is-silver (dismissed):
LGTM. Thx! 🚀

lumirlumir:
> Since this changes the behavior of an existing rule, I’ve updated the PR title to feat. Sorry for the confusion!

This is a behavior change that reports fewer warnings or errors, so it’s fine to leave it as `fix`.

lumirlumir (approved):
Thanks for the contribution. Looks so good to me! 🥳

PR #754, opened 2026-09-16
feat(eslint-markdown): add `consistent-list-indentation` rule

### LINKED ISSUE #726: New Rule: implement `markdownlint`'s `MD005`
### Rule details

Add a rule equivalent to markdownlint's MD005.  The rule should report list item markers that are parsed at the same nesting level but are not consistently aligned. It should also support both left- and right-aligned ordered list markers, matching MD005.  Provide an autofix when the indentation can be corrected safely.

### What type of rule is this?

Enforces a consistent style or user preference

### Example code

```markdown
<!-- Incorrect -->

* Item 1
  * Nested item 1
  * Nested item 2
   * Misaligned item

<!-- Correct -->

* Item 1
  * Nested item 1
  * Nested item 2
  * Nested item 3
```

### Prior Art

- markdownlint: [`MD005` - Inconsistent indentation for list items at the same level](https://github.com/DavidAnson/markdownlint/blob/main/doc/md005.md)

Voyagerroc-Lab:
## Prerequisites checklist

## AI acknowledgment

The rule implementation, tests, and documentation in this pull request were produced with an AI coding agent (model: `zai/zai_auto`). Every change was reviewed and validated locally with the repository's own toolchain (`npm test`, `npm run build`, `npm run lint`).

Implements a new `consistent-list-indentation` rule equivalent to `markdownlint`'s [`MD005` - Inconsistent indentation for list items at the same level](https://github.com/DavidAnson/markdownlint/blob/main/doc/md005.md#md005---inconsistent-indentation-for-list-items-at-the-same-level), as requested in #726.

- `packages/eslint-markdown/src/rules/consistent-list-indentation.ts`: new `layout` rule (autofixable).
- `packages/eslint-markdown/src/rules/consistent-list-indentation.test.ts`: rule tests.
- `packages/eslint-markdown/src/rules/index.ts`: register the new rule.
- `packages/eslint-markdown/src/configs/all.ts`, `packages/eslint-markdown/src/configs/stylistic.ts`: regenerated with `scripts/build-config.js` (the rule is `stylistic`).
- `packages/eslint-markdown/src/index.test-d.ts`: add the new rule/config names to the type assertions.
- `website/docs/rules/consistent-list-indentation.md`: rule documentation.
- `website/docs/get-started/migration.md`: map `MD005` to the new rule.

### Behavior

- Reports list item markers that are parsed at the same nesting level but are not aligned with the first list item of their list. Nested lists are checked independently of their parents.
- Ordered list markers may be either left-aligned (same starting column) or right-aligned (same ending column), matching `MD005`.
- Autofix aligns the leading whitespace before the marker with the first list item of the same list. To keep the fix safe, it is only applied when the leading whitespace consists solely of spaces, so indentation that belongs to another block construct (such as a block quote) is never modified.

All commands were run from the repository root.

- `npm test` - 2875 tests passed, including TypeScript project-reference type checking. This includes the new tests for `consistent-list-indentation` under both `markdown/commonmark` and `markdown/gfm`.
- `npm run build` - package build (including config regeneration) and website build both completed successfully.
- `npm run lint` - `eslint`, `prettier`, `markdownlint`, and `editorconfig-checker` all pass.

The `MD005` ordered-list example taken from `markdownlint`'s own test fixture is included verbatim as a test case and yields the same two violations with the same expected/actual indentation values.

## Related Issues

Fixes #726


packages/eslint-markdown/src/rules/consistent-list-indentation.test.ts
```diff
@@ -0,0 +1,205 @@
+/**
+ * @fileoverview Test for `consistent-list-indentation.ts`.
```
tooth-is-silver, line 2: Every other rule file carries an `@author` tag in its `@fileoverview` block. Add one here to match.

packages/eslint-markdown/src/rules/consistent-list-indentation.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      code: '- [ ] task 1\n- [x] task 2',
+      language: 'markdown/gfm',
+    },
+    {
+      name: 'Consistently indented list inside a block quote',
+      code: '> * item 1\n>   * nested item',
+    },
+  ],
+
+  invalid: [
```
tooth-is-silver, line 73: The ordered-list cases do not cover a marker that is wider than the aligned ones. Please add a case like this:

```ts
 {
   name: 'Right-aligned ordered list followed by a longer marker',
   code: ' 9. a\n10. b\n100. c',
   output: null,
   errors: 1,
 },
```

 It currently throws `RangeError: Invalid count value: -1` rather than reporting.

packages/eslint-markdown/src/rules/consistent-list-indentation.ts
```diff
@@ -0,0 +1,157 @@
...
+    language: 'markdown',
+
+    dialects: ['commonmark', 'gfm'],
+  },
+
+  create(context) {
+    const { sourceCode } = context;
+
+    return {
+      list(node: List) {
```
tooth-is-silver, line 79: The node type is inferred from `RuleModule`, so the annotation is not needed. The `mdast` import can go with it.

```suggestion
      list(node) {
```

```diff
- import type { List } from 'mdast';
```

`npm run test:types` still passes with both removed.

tooth-is-silver (changes requested):

PR #755, opened 2026-09-16
fix(eslint-markdown): remove escaped trailing heading punctuation in `no-trailing-heading-punctuation` rule

lumirlumir:
This pull request introduces a new utility regular expression for detecting escaped trailing backslashes and applies it to improve the handling of shell command continuations and heading punctuation in Markdown linting rules. It also adds comprehensive tests to ensure correct behavior around backslash and punctuation handling.

**Regex utility and rule improvements:**

* Added `escapedTrailingBackslashRegex` to `constants.ts`, which matches an odd-length run of backslashes at the end of a string, and updated its usage in `no-shell-dollar` and `no-trailing-heading-punctuation` rules to more accurately detect line continuations and escaped punctuation. [[1]](diffhunk://#diff-65ea879c6e8285896231eae0cb93bbead08144ad4f60d00076fd929db66aa12dR52-R56) [[2]](diffhunk://#diff-f9237469f0800b5be20e5f3eebc82af3fefbb1cb4bdb6f22171b75e12aae2e2dL12-R12) [[3]](diffhunk://#diff-f9237469f0800b5be20e5f3eebc82af3fefbb1cb4bdb6f22171b75e12aae2e2dL39) [[4]](diffhunk://#diff-f9237469f0800b5be20e5f3eebc82af3fefbb1cb4bdb6f22171b75e12aae2e2dL138-R137) [[5]](diffhunk://#diff-63043c0938ee163419e4e0dd74d17abca71e63a969b9efe84765ad19eeaaa1a1R14) [[6]](diffhunk://#diff-63043c0938ee163419e4e0dd74d17abca71e63a969b9efe84765ad19eeaaa1a1L110-R111) [[7]](diffhunk://#diff-63043c0938ee163419e4e0dd74d17abca71e63a969b9efe84765ad19eeaaa1a1L178-R184) [[8]](diffhunk://#diff-63043c0938ee163419e4e0dd74d17abca71e63a969b9efe84765ad19eeaaa1a1L240-R253)

**Test enhancements:**

* Added unit tests for `escapedTrailingBackslashRegex` in `constants.test.ts` to verify its matching logic for various numbers of trailing backslashes. [[1]](diffhunk://#diff-550cca8abc644491d2f627ebd8e979d41a6425839f4a3d9c41daaff7ee3c86ffR14) [[2]](diffhunk://#diff-550cca8abc644491d2f627ebd8e979d41a6425839f4a3d9c41daaff7ee3c86ffR65-R86)
* Expanded test cases in `no-trailing-heading-punctuation.test.ts` to cover headings with different combinations of trailing backslashes and punctuation, ensuring correct linting and autofix behavior.

These changes increase the robustness and correctness of Markdown linting around shell command continuations and heading punctuation, especially in edge cases involving backslashes.


packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts
```diff
@@ -175,10 +176,12 @@ export default {
         const lastChildText = sourceCode.getText(lastChildNode);
         const match = trailingPunctuationRegex.exec(lastChildText);
 
-        if (!match) {
+        if (!match || !match.groups) {
           return;
         }
 
+        const { leadingWhitespace = '' } = match.groups;
```
tooth-is-silver: It may be worth checking whether `match.groups` can actually be `undefined` here.
TypeScript defines groups as optional in general, but this regex always contains a named capture group.
Also, `leadingWhitespace` cannot be `undefined` because `*` captures an empty string when no whitespace is present.

lumirlumir: > It may be worth checking whether match.groups can actually be undefined here.
TypeScript defines groups as optional in general, but this regex always contains a named capture group.

Yep, this check is about personal preference, as I prefer not to use the non-null assertion operator unless it’s truly necessary.

https://github.com/eslint-markdown/eslint-markdown/blob/85cfba2b2fd2b0df11c52ac212d089829286a870/packages/eslint-markdown/src/rules/require-heading-id.ts#L220

---

> Also, leadingWhitespace cannot be undefined because * captures an empty string when no whitespace is present.

👍 updated it in https://github.com/eslint-markdown/eslint-markdown/pull/755/commits/fff227cc3252e816536bc3d74d1011ad63757cdd

packages/eslint-markdown/src/core/constants.test.ts
```diff
@@ -61,6 +62,28 @@ describe('constants', () => {
...
+      });
+
+      it('should not match four trailing backslashes', () => {
+        assert.notMatch(`text${'\\'.repeat(4)}`, escapedTrailingBackslashRegex);
+      });
+
+      it('should not match a backslash before the end of a line', () => {
+        assert.notMatch('text\\more', escapedTrailingBackslashRegex);
+      });
+    });
```
tooth-is-silver, line 85: nit: the JSDoc describes this as an _odd-length_ run, but the it descriptions only give counts, so the odd and even split is not visible. Could they say that more directly?

lumirlumir: 👍 Updated it in https://github.com/eslint-markdown/eslint-markdown/pull/755/commits/3d96840cf40ed3238c27f5091d9af4e9dbde8a46

tooth-is-silver (approved):
LGTM 🚀

PR #756, opened 2026-09-16
feat(eslint-markdown): add autofix and suggestion to `consistent-code-style`

### LINKED ISSUE #734: Rule Change: add autofix and suggestion to `consistent-code-style` rule
Add autofix and suggestion to `consistent-code-style` rule.

https://github.com/eslint-markdown/eslint-markdown/blob/85cfba2b2fd2b0df11c52ac212d089829286a870/packages/eslint-markdown/src/rules/consistent-code-style.ts#L6-L14

https://github.com/eslint-markdown/eslint-markdown/blob/85cfba2b2fd2b0df11c52ac212d089829286a870/packages/eslint-markdown/src/rules/consistent-code-style.ts#L72-L74

Voyagerroc-Lab:
## Prerequisites checklist

## AI acknowledgment

This pull request was written by an AI agent (AutoClaw, model `zai_auto`): the rule change, the new test cases, and the `website/` docs. Everything claimed under "How did you test this change?" comes from commands that were actually run against this branch.

`consistent-code-style` already knew which style a code block should have, but it could only say so: both `fixable: 'code'` and `hasSuggestions: true` were left as `// TODO`, so `--fix` did nothing for this rule and a consumer had to reflow every reported block by hand.

The file already carried the safety analysis for the conversions it wanted to offer, so this pull request implements exactly that analysis and leaves the destinations it calls out as unsafe reported without a fix.

`packages/eslint-markdown/src/rules/consistent-code-style.ts`

- Declares `fixable: 'code'` and `hasSuggestions: true`, and adds a `suggestIndent` message for the suggestion described below.
- Adds helpers that convert a code block from one style to another, and one entry point (`getStyleConversion`) that decides how a block reaches the expected style. The decision follows the note at the top of the file:
  - `fence-backtick` ⇄ `fence-tilde` — autofix. Only the two fence runs are rewritten, and with runs of the same length, so a container prefix (a blockquote's `>`, a list item's indentation) is carried over untouched and the info string is kept.
  - `indent` → `fence-backtick` / `fence-tilde` — autofix, at the top level only.
  - `fence-backtick` / `fence-tilde` → `indent` — autofix when the fence carries no info string, otherwise a suggestion, since the info string is dropped and cannot be recovered. The conversion is only offered where the block already sits between blank lines, because unlike a fence an indented code block cannot interrupt a paragraph and is ended by a blank line.
  - No fix at all when the conversion has no faithful form: a content line that would close the new fence early, a backtick in the info string of a fence that is becoming a backtick fence, an empty block (an indented code block cannot be empty), a block whose container prefix the node range does not carry, or nothing but a fence character to convert.
- The reported location for the `style` message is unchanged, and so is the behaviour of the `blankLineAbove` / `blankLineBelow` messages, which are still reported without a fix. Their conversions would have to insert blank lines that carry a container prefix (the blank lines in a blockquote are written `>`), which is a separate problem from the one this rule's safety note is about.

`packages/eslint-markdown/src/rules/consistent-code-style.test.ts`

- Existing `invalid` cases for the `style` message now assert `output` or `suggestions`, and new cases cover the cases above: blockquoted and nested blockquoted fences, a fence longer than the minimum, a backtick fence inside a tilde block, an indented block containing a backtick fence, a backtick in an info string, a fence next to a paragraph, an unclosed fence, an empty fence, and tab-indented code.

`website/docs/rules/consistent-code-style.md`

- Adds the `## Fix` and `## Suggestion` sections that the other fixable rules document, matching `consistent-heading-style`.

```sh
npm run test:types   # tsc -b, clean
npm run test:unit    # 39 files, 2846 tests passed
npx vitest run packages/eslint-markdown/src/rules/consistent-code-style.test.ts
                     # 149 passed (was 127)
npx eslint packages/eslint-markdown/src/rules/consistent-code-style.ts \
             packages/eslint-markdown/src/rules/consistent-code-style.test.ts
                     # no findings
npx prettier --check packages/eslint-markdown/src/rules/consistent-code-style.ts \
                     packages/eslint-markdown/src/rules/consistent-code-style.test.ts \
                     website/docs/rules/consistent-code-style.md
                     # all files use Prettier code style
npx markdownlint website/docs/rules/consistent-code-style.md
                     # no findings
```

The tests are not vacuous: checking out `main`'s version of the rule while keeping the updated test file fails 36 of the 149 cases, which are exactly the conversions the rule could not perform before.

The conversions are also checked for what they must *not* change. `fixer.replaceText(node, ...)` is only reached with text that reproduces every character outside the fence runs, which is why the blockquoted cases assert that the `> ` prefixes survive.

## Related Issues

Closes #734

- **The `blankLineAbove` / `blankLineBelow` messages stay unfixed.** They are the other two messages this rule reports, so `hasSuggestions` / `fixable` could be read as covering them. Making them fixable means inserting blank lines that may need a container prefix, so it looked like a separate change; happy to add it here or in a follow-up if you would rather it landed together.
- **The name `suggestIndent` and its wording** are a guess at the convention `consistent-heading-style` sets with `suggestAtxToSetext`.
- **Whether the suggestion should exist at all.** A suggestion is still a conversion that loses an info string. It is only offered when the fence actually carries one, so applying it is always a deliberate choice, but reviewing the call is worthwhile.


lumirlumir:
Thanks for the PR, but we don't accept PRs that were primarily generated by AI.

PR #759, opened 2026-09-16
feat(eslint-markdown): add `style` option to `no-emoji`

### LINKED ISSUE #710: Rule Change: add `style` option to `no-emoji`
## Description

Add a `style` option to the `no-emoji` rule so it can target raw Unicode emoji (`😃`), `:shortcode:` style gemoji (`:smiley:`), or both. Currently the rule only detects raw Unicode emoji.

## Motivation

Some projects want the opposite convention (flag shortcodes, allow raw emoji), or want both forms disallowed at once. There's currently no way to configure this.

## Proposed Solution

- Add `style: ('emoji' | 'gemoji')[]`, default `['emoji']` (no breaking change).
- When `'gemoji'` is included, match using the existing `gemojiRegex` from [src/core/constants.ts](https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/core/constants.ts#L57) .
- Both styles can be enabled together and are checked independently.
- `allow` continues to work the same across all enabled styles.
- No auto-fix, consistent with current behavior.

## Additional Context

- Rule source: https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/rules/no-emoji.ts
- gemojiRegex: https://github.com/lumirlumir/npm-eslint-markdown/blob/main/packages/eslint-markdown/src/core/constants.ts#L57
- Docs: https://eslint-markdown.lumir.page/docs/rules/no-emoji

ParkTjgus:
## Prerequisites checklist

## AI acknowledgment

Claude Code (Claude Opus 5) wrote the tests and documentation according to the design I provided. I reviewed every line of the generated code and documentation.

`no-emoji` currently only disallows raw emojis (e.g. `😃`). Some projects want to allow raw emojis and disallow emoji shortcodes (e.g. `:smiley:`). There was no way to configure this.

- Added a `style` option (`'emoji' | 'gemoji'`) to `no-emoji`.
  - `'emoji'`(default): disallows raw emojis.
  - `'gemoji'`: disallows emoji shortcodes, matched with the existing `gemojiRegex` from `src/core/constants.ts`.
  - The default is `'emoji'`, so this is not a breaking change.
- Following the review feedback, `style` is a single value (union type) rather than an array, so only one style can be disallowed at a time in this version.
- `allow` works with the selected style: list raw emojis (e.g. `'😃'`) for `'emoji'` and emoji shortcodes (e.g. `':smiley:'`) for `'gemoji'`.
- No auto-fix was added, consistent with the current behavior.
- Updated `website/docs/rules/no-emoji.md`:
  - Added incorrect/correct examples for `{ style: 'gemoji' }` and a correct example for `{ style: 'gemoji', allow: [':unicorn:'] }`.
  - Following the `consistent-*-style` rule docs, the default value (`'emoji'`) is covered by the `Default` examples and is not repeated as a separate option example.
  - Documented the `style` option and updated the `allow` description.
  - Added a limitation note: the `'gemoji'` style matches the `:name:` pattern rather than a list of real emoji names, so text like `:min:` in `hour:min:sec` is also reported.

- Added test cases to `packages/eslint-markdown/src/rules/no-emoji.test.ts`:
  - Each style disallows its own format and allows the other format.
  - `allow` works with each style.
  - Adjacent shortcodes and multi-line shortcodes are reported at the correct locations.
- Ran every `eslint-check` example in `website/docs/rules/no-emoji.md` against the rule and confirmed that the incorrect examples report errors and the correct examples do not.
- Confirmed that invalid `style` values (e.g. `[]`, `['emoji']`, `'foo'`) are rejected by the schema.
- Ran the following commands from the repository root:
  - `npm run test`: 39 test files, 2844 tests passed
  - `npm run build`: passed
  - `npm run lint`: passed

## Related Issues

fixes #710

- Emojis inside inline code and code blocks are not reported. As before, the rule only checks `text` nodes, so this applies to both raw emojis and emoji shortcodes.
- The `'gemoji'` style reuses `gemojiRegex`, which matches the `:name:` pattern rather than a list of real emoji names. As a result, text that is not an emoji can be reported. For example, `:min:` in `hour:min:sec` is reported. I documented this in the "Limitations" section instead of changing the shared regex, because it is also used by `no-trailing-heading-punctuation`.


packages/eslint-markdown/src/rules/no-emoji.ts
```diff
@@ -22,10 +22,18 @@ type RuleOptions = [
...
      */
     allow: string[];
+    /**
+     * Specifies the style of emojis to disallow.
+     *
+     * - `'emoji'`: Raw Unicode emojis (`😃`).
+     * - `'gemoji'`: Shortcode style emojis (`:smiley:`).
+     * @default ['emoji']
+     */
+    style: ('emoji' | 'gemoji')[];
```
lumirlumir:
```suggestion
     * @default ['emoji']
     */
    style: 'emoji' | 'gemoji';
```

Can we use union style for this case rather than using array one?

I want the first version of this feature to be an union style option.

packages/eslint-markdown/src/rules/no-emoji.ts
```diff
@@ -62,6 +71,14 @@ export default {
             },
             uniqueItems: true,
           },
+          style: {
+            type: 'array',
+            items: {
+              enum: ['emoji', 'gemoji'],
+            },
+            minItems: 1,
+            uniqueItems: true,
```
lumirlumir: Same here. I hope this option to be union type 🙂

lumirlumir (changes requested):
Thanks for the PR!

I've actually misread the issue, and my original intention was to make this option as `style: 'emoji' | 'gemoji'` type. I've left a few comments about them.

packages/eslint-markdown/src/rules/no-emoji.ts
ParkTjgus: Thanks for the quick review. With a union type, we wouldn’t be able to restrict both values simultaneously. Would you prefer to have the restriction apply to only one of the emoji types in the first version?

lumirlumir: Yep, that's my first intention of this change, so I think it's fine.

packages/eslint-markdown/src/rules/no-emoji.ts
ParkTjgus: Thank you for the review! I’ve addressed the feedback in commit [4ea32f7](https://github.com/eslint-markdown/eslint-markdown/pull/759/changes/4ea32f77e233e546bc25e5733f4783069b96c214). Would you mind reviewing it again?

lumirlumir (approved):
Thanks for the contribution!

PR #763, opened 2026-09-17
refactor(eslint-markdown): add ASCII and non-ASCII punctuation constants with question mark variations

lumirlumir:
This pull request refactors how punctuation character sets are defined and used in the `eslint-markdown` package. It introduces named constants for ASCII and non-ASCII punctuation (with and without question marks), updates the main punctuation arrays to use these new constants, and ensures consistent usage across the codebase and tests. Additionally, the `no-double-punctuation` rule now dynamically uses the ASCII punctuation set, making it easier to maintain and extend.

**Punctuation character set refactoring:**

* Introduced new constants in `constants.ts`: `asciiPunctuation`, `asciiPunctuationWithQuestionMark`, `nonAsciiPunctuation`, and `nonAsciiPunctuationWithQuestionMark`, and refactored `punctuation` and `punctuationWithQuestionMark` to be derived from these.
* Updated related tests in `constants.test.ts` to check for the new constants and their correct values. [[1]](diffhunk://#diff-550cca8abc644491d2f627ebd8e979d41a6425839f4a3d9c41daaff7ee3c86ffR12-R15) [[2]](diffhunk://#diff-550cca8abc644491d2f627ebd8e979d41a6425839f4a3d9c41daaff7ee3c86ffR44-R81) [[3]](diffhunk://#diff-550cca8abc644491d2f627ebd8e979d41a6425839f4a3d9c41daaff7ee3c86ffL59-R114)

**Rule improvements:**

* Updated `no-double-punctuation.ts` to import and use `asciiPunctuationWithQuestionMark` for building the double punctuation regex and schema validation, ensuring the rule is always in sync with the defined punctuation set. [[1]](diffhunk://#diff-ae55d05e7c659eb90dac6e130409d6e4eee25f71971c8c999888d37bf4e4e839L10-R10) [[2]](diffhunk://#diff-ae55d05e7c659eb90dac6e130409d6e4eee25f71971c8c999888d37bf4e4e839R38-R47) [[3]](diffhunk://#diff-ae55d05e7c659eb90dac6e130409d6e4eee25f71971c8c999888d37bf4e4e839L73-R78)


packages/eslint-markdown/src/core/constants.ts
```diff
@@ -18,30 +18,45 @@ export const URL_RULE_DOCS = (ruleName = ''): string =>
...
 /**
  * This pattern is based on the punctuation list used by `markdownlint` and `remark-lint`.
  * @see https://github.com/DavidAnson/markdownlint/blob/v0.41.1/helpers/helpers.cjs#L41
  * @see https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-heading-punctuation#parameters
  */
-export const punctuationWithQuestionMark = [...punctuation, '?', '？'] as const;
+export const punctuationWithQuestionMark = [
+  ...asciiPunctuationWithQuestionMark,
+  ...nonAsciiPunctuationWithQuestionMark,
+] as const;
```
tooth-is-silver, line 59: `nonAsciiPunctuationWithQuestionMark` is only used to build this array, so it seems unnecessary as a separate constant.
Would it be simpler to spread punctuation and add the question marks directly?

```ts
export const punctuationWithQuestionMark = [...punctuation, '?', '？'] as const;
```

lumirlumir: Yep, actually this violates YAGNI, but this is for a future usage and for consistency with other helper array, so I'd like to keep it in this version. 🙂 Thanks!

tooth-is-silver (approved):
LGTM 🚀

PR #764, opened 2026-09-17
feat(eslint-markdown): add `punctuation` option to `no-double-punctuation`

### LINKED ISSUE #762: Rule Change: add `punctuation` option to `no-double-punctuation` rule
Add a `punctuation` option to `no-double-punctuation`, following the option design used by `no-trailing-heading-punctuation`.

## Motivation

`no-double-punctuation` currently always detects pairs made from the built-in ASCII punctuation set (including `?`). The existing `allow` option can exempt individual two-character combinations, but it cannot narrow or replace the set of punctuation characters the rule examines. Projects that intentionally use a smaller punctuation set must enumerate every permitted combination in `allow`.

A `punctuation` option would make the rule configurable at the same level as `no-trailing-heading-punctuation` while preserving the current behavior by default.

## Proposed option

Add a `punctuation: string[]` property to the rule options.

https://github.com/eslint-markdown/eslint-markdown/blob/2d58f54127ff3a1dda55b74119280bdbe3c91a6a/packages/eslint-markdown/src/rules/no-trailing-heading-punctuation.ts#L90-L94

```js
{
  'md/no-double-punctuation': ['error', {
    punctuation: ['.', '!', '?'],
    allow: ['?!', '!!'],
  }],
}
```

- Default: the current `asciiPunctuationWithQuestionMark` list, so existing configurations remain unchanged.
- Validation: an array of unique one-character strings with at least one item, matching the `punctuation` schema of `no-trailing-heading-punctuation`.
- Detection: build the double-punctuation regular expression from the configured `punctuation` list instead of the fixed constant.
- `allow`: continue to apply after detection, and validate each allowed two-character pattern against the configured punctuation list. This keeps `allow` focused on intentional combinations such as `?!`.

## Expected behavior

With `{ punctuation: ['.', '!'] }`:

- `Hello..` and `Hello!!` are reported.
- `Hello?!` is not reported because `?` is outside the configured set.
- `allow: ['!!']` exempts `Hello!!`.
- `allow: ['??']` should be rejected by option validation because `?` is not configured.

## Implementation and test notes

- Reuse `escapeStringRegexp` when constructing the configured character class, as `no-trailing-heading-punctuation` does.
- Keep the existing autofix behavior for duplicate characters and the existing suggestions for mixed characters.
- Add schema/default-option tests and cases covering custom punctuation, `allow` interaction, and backward-compatible defaults.

tunglambk:
## Prerequisites checklist

## AI acknowledgment

DeepSeek Harness (deepseek-v4.1-flash) wrote the implementation, tests, and documentation, and ran the checks below against the final commit.

`no-double-punctuation` always detects pairs from the built-in ASCII punctuation set, and `allow` can only exempt specific two-character combinations. A project that intentionally uses a smaller set has to enumerate every permitted pair, so this adds a `punctuation` option that can narrow or replace the detected set, following the option design already used by `no-trailing-heading-punctuation`.

- Added `punctuation: string[]` to `no-double-punctuation`, defaulting to `['.', ',', ';', ':', '!', '?']`. The double punctuation regular expression is built from the configured set with `escapeStringRegexp`, and the schema requires at least one unique one-character item.
- The regex is compiled with the `u` flag, so a configured astral character is matched as one character instead of two UTF-16 code units.
- `allow` still applies after detection, and each pattern is now validated against the configured set. A pattern that uses a character outside `punctuation` makes the rule fail to load with a clear message instead of being silently ignored.
- The existing autofix for duplicate characters and the suggestions for mixed pairs are unchanged, and no option supplied still behaves exactly as before.

**Scope:** `packages/eslint-markdown/src/rules/no-double-punctuation.ts`, `packages/eslint-markdown/src/rules/no-double-punctuation.test.ts`, `website/docs/rules/no-double-punctuation.md`

**Validation** (all from the repository root):

- `npm install` — passed
- `npm run test` — passed, 39 test files and 2882 tests
- `npm run build` — passed
- `npm run lint` — passed (eslint, prettier, markdownlint, editorconfig-checker; no download failure this run)

I ran the new cases against the pre-change rule first: 20 failed, including the option cases and the `allow` validation, and they pass after the change.

The tests cover the default set, narrowing and replacing the set, non-ASCII and astral characters, the `allow` interaction, and the rejection of an `allow` pattern that uses an unconfigured character. The `eslint-check` examples in `website/docs/rules/no-double-punctuation.md` were run against the built rule and report as documented.

## Related Issues

Fixes #762

**Notes:** the `allow` patterns are checked against `punctuation` when the rule is created rather than in `meta.schema`, because JSON Schema can't reference a sibling option property and ESLint's Ajv instance doesn't enable `$data`. If you'd rather have an `allow` entry outside the configured set ignored instead of rejected, that's a small change. `allow` entries are also length-checked in UTF-16 code units by the schema, so an astral pair such as `allow: ['😀😀']` is rejected even though the regex now matches it — that's deliberately out of scope for this change.


tunglambk:
Reproduced. Without the `u` flag, `punctuation: ['😀']` treats the emoji as two UTF-16 code units, so `Foo😀` was reported as double punctuation while `Foo😀😀` was not reported at all. I did not see the fixer actually corrupt the source, though: the replacement text comes from code-point destructuring, so the no-op fix left the emoji intact. The matching was definitely mis-split, so I added the `u` flag where the regex is built. ASCII output is unchanged, and two new tests pin that a single astral character is not reported and that a pair is reported and fixed to one. The docs now say characters are matched by code point.

PR #766, opened 2026-09-17
fix(eslint-markdown): handle case-insensitive `aria-hidden` values in `require-link-title` and `require-image-title`

lumirlumir:
This pull request improves the handling of the `aria-hidden="true"` attribute in both the `require-image-title` and `require-link-title` rules, making the check case-insensitive and updating related tests and documentation accordingly.

**Improvements to rule logic:**

* Updated the logic in `require-image-title.ts` and `require-link-title.ts` to treat the `aria-hidden="true"` attribute as case-insensitive by comparing the lowercased value, so values like `"TRUE"` or `"TrUe"` are now correctly recognized. [[1]](diffhunk://#diff-a271aca9751c39eb00354b8b0b9a685cb31fc4bb311a208bfc97e0ce51fbe673L134-R134) [[2]](diffhunk://#diff-2f5c0a0cf60c4bc914217f9dfcfda04bf1dd9235ecd21661f2091086de61add9L129-R129)

**Test enhancements:**

* Added test cases to `require-image-title.test.ts` and `require-link-title.test.ts` to verify that mixed- and upper-case variants of `aria-hidden="true"` are handled correctly, and to ensure that only exact matches (not substrings or values with line feeds) are ignored. [[1]](diffhunk://#diff-6135f3b3dba0adfe06f39c7745db9b71dcb0a0cd51412d6e78c4d9c8b0fb5707R107-R114) [[2]](diffhunk://#diff-7a667adf7f0da3c2abf24fd14aeedb6f51314a47a717d671ac13992c4e55b851R110-R117) [[3]](diffhunk://#diff-6135f3b3dba0adfe06f39c7745db9b71dcb0a0cd51412d6e78c4d9c8b0fb5707R408-R433) [[4]](diffhunk://#diff-7a667adf7f0da3c2abf24fd14aeedb6f51314a47a717d671ac13992c4e55b851R385-R410)

**Documentation updates:**

* Updated the documentation for both rules to clarify that `aria-hidden="true"` is now treated in a case-insensitive manner. [[1]](diffhunk://#diff-78f27e3642250175c76807ad1205285437bedfbb4c10fefde8033999346446edL16-R16) [[2]](diffhunk://#diff-6f55d0761a208353e3de1edfe7d1f17111cfb29048326e03d43700bbf71671d5L16-R16)


packages/eslint-markdown/src/rules/require-link-title.ts
```diff
@@ -126,7 +126,7 @@ export default {
               break;
             }
 
-            if (name === 'aria-hidden' && value === 'true') {
+            if (name === 'aria-hidden' && /^true$/iu.test(value)) {
```
lumirlumir: This is a small regex util, so I intentionally didn't promote it into shared constants.

packages/eslint-markdown/src/rules/require-image-title.test.ts
```diff
@@ -104,6 +104,14 @@ ruleTester('require-image-title', rule, {
...
       code: '<img src="https://example.com/image.jpg" aria-hidden="true">',
     },
+    {
+      name: 'Html node with aria-hidden attribute set to mixed-case true',
+      code: '<img src="https://example.com/image.jpg" aria-hidden="TrUe">',
+    },
+    {
+      name: 'Html node with aria-hidden attribute set to uppercase true',
+      code: '<img src="https://example.com/image.jpg" aria-hidden="TRUE">',
+    },
```
tooth-is-silver, line 114: Some browsers treat `aria-hidden` values as case-sensitive. [W3C’s ACT](https://www.w3.org/WAI/standards-guidelines/act/rules/6cfa84/#accessibility-support)
Therefore, `"TRUE"` and `"TrUe"` may not hide the element from assistive technology.
I think only the exact `"true"` value should skip this rule, and the mixed-case examples should remain invalid.

tooth-is-silver: I confirmed that we agreed to follow `@eslint/markdown`’s behavior. Please disregard this comment. 👍

tooth-is-silver (approved):
LGTM 🚀

lumirlumir:
@tooth-is-silver 

The prior arts from `@eslint/markdown`: 

- https://github.com/eslint/markdown/pull/465 
- https://github.com/eslint/markdown/blob/main/src/rules/no-html.js#L28
- https://github.com/eslint/markdown/blob/main/src/rules/no-missing-link-fragments.js#L31-L36
- https://github.com/eslint/markdown/blob/main/src/rules/no-multiple-h1.js#L28
- https://github.com/eslint/markdown/blob/main/src/rules/require-alt-text.js#L28-L30

PR #769, opened 2026-09-17
fix(eslint-markdown): ignore punctuation after gemoji and HTML entities in `no-double-punctuation`

### LINKED ISSUE #758: Bug: `no-double-punctuation` reports punctuation after gemoji and HTML entities
### Environment

ESLint version: 9.39.5
@eslint/markdown version: 8.0.3
eslint-markdown version: 0.17.0
Node version: 24.14.0
npm version: 11.9.0
Operating System: macOS 26.5

### Which language are you using?

commonmark & gfm

### What did you do?

<details>
<summary>Configuration</summary>

```js
import { defineConfig } from 'eslint/config';
import markdown from '@eslint/markdown';
import md from 'eslint-markdown';

export default defineConfig([
  {
    files: ['**/*.md'],
    plugins: {
      markdown,
      md,
    },
    language: 'markdown/commonmark',
    rules: {
      'md/no-double-punctuation': 'error',
    },
  },
]);
```

</details>

```md
This is :smile:.
Copyright &copy;.
```

Run the following command:

```sh
npm run lint
```

The same behavior occurs with both CommonMark and GFM.
The same behavior occurs with both CommonMark and GFM.

### What did you expect to happen?

No errors should be reported.

The colon is part of the gemoji, and the semicolon is part of the HTML entity. They should not be treated as consecutive punctuation with the period that follows.

### What actually happened?

The rule reports `:.` and `;.` as double punctuation:

Image

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/stackblitz-starters-s1olntc4?file=eslint.config.js,commonmark.md

### Participation

- [x] I am willing to submit a pull request for this issue.

### AI acknowledgment

- [ ] No - I did not use AI to generate this issue report.
- [x] Yes - minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes - substantive assistance (AI-generated non-trivial portions of code, tests, or documentation).

### Additional comments

_No response_

tooth-is-silver:
## Prerequisites checklist

## AI acknowledgment

- [x] Yes - minimal assistance (e.g., IDE autocomplete, small code completions, grammar fixes).
- [ ] Yes - substantive assistance (AI-generated non-trivial portions of code, tests, or documentation).

`no-double-punctuation` reports the punctuation that follows a gemoji or an HTML entity, because the semicolon that closes them counts as the first of the pair. `Copyright &copy;! All rights reserved.` and `Nice work :+1:!` are both reported, even though neither contains double punctuation.

This makes the rule skip a match when the character before it closes a gemoji or an HTML entity.

Added `trailingGemojiRegex` and `trailingHtmlEntityRegex` to `src/core/constants.ts`, anchored to the end of the text.

Updated `no-double-punctuation` to test the text up to and including the first punctuation character against both, and to skip the match when either one matches.

`no-trailing-heading-punctuation` built its own end-anchored gemoji expression inline. It now imports the shared one instead.

I added the following test cases.

Valid:
- Punctuation after a gemoji, such as :+1:!
- Punctuation after an HTML entity, such as &copy;!

Invalid:
- Double punctuation that only looks like a gemoji is still reported
- Double punctuation that only looks like an HTML entity is still reported

The two new constants are covered in `src/core/constants.test.ts`, including the case where a gemoji or an entity is followed by more text rather than ending it.

- `npm run test`
- `npm run lint`

## Related Issues

fixes #758


packages/eslint-markdown/src/core/constants.ts
```diff
@@ -79,3 +79,31 @@ export const gemojiRegex =
...
+// --------------------------------------------------------------------------------
+// #region regex
+
+/**
+ * Regular expression for identifying a GitHub emoji code at the end of a line.
+ */
+export const trailingGemojiRegex = new RegExp(
+  `${gemojiRegex.source}$`,
+  gemojiRegex.flags,
+);
```
lumirlumir: I’m not sure this regex belongs here. I’d prefer to derive it from the original regex in the rule itself rather than maintain another similar version here for now.

We can consider moving it into a shared utility later if more than three or four patterns are used across the rules.

packages/eslint-markdown/src/core/constants.ts
```diff
@@ -79,3 +79,31 @@ export const gemojiRegex =
 
 // #endregion regex
 // --------------------------------------------------------------------------------
+
+// --------------------------------------------------------------------------------
+// #region regex
```
lumirlumir:
```suggestion

```

This region marker is for VS Code and we don’t need a separate one. Can we consolidate it into the existing one?

packages/eslint-markdown/src/core/constants.ts
```diff
@@ -79,3 +79,31 @@ export const gemojiRegex =
...
+export const trailingGemojiRegex = new RegExp(
+  `${gemojiRegex.source}$`,
+  gemojiRegex.flags,
+);
+
+// #endregion regex
+// --------------------------------------------------------------------------------
+
+// --------------------------------------------------------------------------------
+// #region regex
```
lumirlumir:
```suggestion

```

packages/eslint-markdown/src/core/constants.ts
```diff
@@ -79,3 +79,31 @@ export const gemojiRegex =
...
+/**
+ * Regular expression for identifying an HTML entity at the end of a line.
+ * - NOTE: These patterns are based on the `markdownlint`.
+ * @see https://github.com/DavidAnson/markdownlint/blob/v0.41.1/helpers/helpers.cjs#L32-L34
+ */
+export const trailingHtmlEntityRegex =
+  /&(?:#\d+|#[xX][\da-fA-F]+|[a-zA-Z]{2,31}|blk\d{2}|emsp1[34]|frac\d{2}|sup\d|there4);$/;
+
+// #endregion regex
+// --------------------------------------------------------------------------------
```
lumirlumir, line 89:
```suggestion
```

packages/eslint-markdown/src/core/constants.ts
```diff
@@ -79,3 +79,31 @@ export const gemojiRegex =
...
+// #region regex
+
+/**
+ * Regular expression for identifying an HTML entity at the end of a line.
+ * - NOTE: These patterns are based on the `markdownlint`.
+ * @see https://github.com/DavidAnson/markdownlint/blob/v0.41.1/helpers/helpers.cjs#L32-L34
+ */
+export const trailingHtmlEntityRegex =
+  /&(?:#\d+|#[xX][\da-fA-F]+|[a-zA-Z]{2,31}|blk\d{2}|emsp1[34]|frac\d{2}|sup\d|there4);$/;
```
lumirlumir:
```suggestion
export const htmlEntityRegex =
  /&(?:#\d+|#[xX][\da-fA-F]+|[a-zA-Z]{2,31}|blk\d{2}|emsp1[34]|frac\d{2}|sup\d|there4);/;
```

This pattern could also serve as a general-purpose regex in the future, so I think it would be better to decouple it from the `$` regex symbol, as we do with `gemojiRegex`.

It would also be better to derive it in the rule itself using `new RegExp()` and append `$` there.

lumirlumir (changes requested):

packages/eslint-markdown/src/core/constants.ts
tooth-is-silver: Updated in [06b9f37](https://github.com/eslint-markdown/eslint-markdown/pull/769/commits/06b9f376fed8992957067b81c80f927cb62997ee)

packages/eslint-markdown/src/core/constants.ts
tooth-is-silver: Updated in [d45ab68](https://github.com/eslint-markdown/eslint-markdown/pull/769/commits/d45ab6800cfac8d76ab372982bfc1fcbb25c6d45)

packages/eslint-markdown/src/core/constants.ts
tooth-is-silver: Thank you for the feedback. I’ve updated it to derive the regex within each rule.

packages/eslint-markdown/src/core/constants.test.ts
```diff
@@ -248,5 +249,17 @@ describe('constants', () => {
         });
       });
     });
+
+    describe('htmlEntityRegex', () => {
```
lumirlumir, line 253: Could you add more comprehensive tests for `htmlEntityRegex` here?

As you can see with `gemojiRegex`, we have many tests to help prevent potential regressions.

It would be helpful to add test cases for `#\d+`, `#[xX][\da-fA-F]+`, `[a-zA-Z]{2,31}`, `blk\d{2}`, `emsp1[34]`, and all the other patterns separated by the `|` marker.

lumirlumir (changes requested):

packages/eslint-markdown/src/core/constants.test.ts
tooth-is-silver: Updated id [f67e7c2](https://github.com/eslint-markdown/eslint-markdown/pull/769/commits/f67e7c2ca06c68254b4bac42ef703a5856956842)

tooth-is-silver:
@lumirlumir 
I added two test cases for HTML entities and reorganized the existing tests for clarity. [f67e7c2](https://github.com/eslint-markdown/eslint-markdown/pull/769/commits/f67e7c2ca06c68254b4bac42ef703a5856956842)

packages/eslint-markdown/src/rules/no-double-punctuation.test.ts
lumirlumir: Can we also add the following `valid` test case here as well?

```md
:banana::tada::confetti_ball::partying_face:
```

packages/eslint-markdown/src/rules/no-double-punctuation.test.ts
```diff
@@ -61,6 +61,24 @@ ruleTester('no-double-punctuation', rule, {
...
     },
+    // gemoji
+    {
+      name: 'gemoji',
+      code: 'This is :smile:.',
+    },
+    {
+      name: 'gemoji',
+      code: 'This is :smile:. Next sentence.',
+    },
```
lumirlumir:
```suggestion
    },
    {
      name: 'gemoji - 3',
      code: 'hi.:tada:',
    },
```

There’s still a bug when a gemoji or HTML entity appears later in the text.

I think we can first mask the gemoji or HTML entity with whitespace and then check for consecutive punctuation.

For example:

```
hi.:tada:bye
   ^^^^^^
```

would become:

```
hi.      bye
   ^^^^^^
```

We could then apply the existing logic so that gemoji and HTML entities don’t interfere with the consecutive-punctuation check.

lumirlumir (changes requested):

packages/eslint-markdown/src/core/constants.ts
```diff
@@ -77,5 +77,13 @@ export const escapedTrailingBackslashRegex = /(?<=(?<!\\)(?:\\{2})*)\\$/u;
 export const gemojiRegex =
   /:(?:[abmovx]|[-+]1|100|1234|(?:1st|2nd|3rd)_place_medal|8ball|clock\d{1,4}|e-mail|non-potable_water|o2|t-rex|u5272|u5408|u55b6|u6307|u6708|u6709|u6e80|u7121|u7533|u7981|u7a7a|[a-z]{2,15}2?|[a-z]{1,14}(?:_[a-z\d]{1,16})+):/;
 
+/**
+ * Regular expression for identifying an HTML entity.
+ * - NOTE: This pattern is based on `markdownlint`.
+ * @see https://github.com/DavidAnson/markdownlint/blob/v0.41.1/helpers/helpers.cjs#L32-L34
+ */
+export const htmlEntityRegex =
```
lumirlumir: `sourceCode.getText()` preserves Markdown backslashes, so `htmlEntityRegex` currently recognizes `\&copy;` as an HTML entity even though GitHub renders it literally as `&copy;`.

This causes false negatives in both rules:

- `no-double-punctuation`: `\&copy;.` incorrectly ignores `;.`
- `no-trailing-heading-punctuation`: `# \&copy;` incorrectly ignores the trailing `;`

Since both rules use `htmlEntityRegex` against raw Markdown source, could we make the shared regex itself escape-aware?

```ts
export const htmlEntityRegex =
  /(?<=(?<!\\)(?:\\{2})*)&(?:#\d+|#[xX][\da-fA-F]+|[a-zA-Z]{2,31}|blk\d{2}|emsp1[34]|frac\d{2}|sup\d|there4);/;
```

This should apply only to HTML entities. GitHub still renders `\:smile:` as an emoji, so `gemojiRegex` should remain unchanged.

Could you also add:

- Odd/even backslash coverage to `constants.test.ts`
- A `\&copy;.` regression test for `no-double-punctuation`
- A `# \&copy;` regression test for `no-trailing-heading-punctuation`

PR #772, opened 2026-09-17
chore(*): remove `sync-client`

lumirlumir:
These files were used to sync configuration files across the repository I've maintained, but this repo is now separated, it's no longer needed.


lumirlumir:
Ah, I think we can retain it. So closing.

